PR #11, opened 2018-06-18
Finished processing csv files

MartinBentley:


src/ManageCourses.Domain/ManageCourses.Domain.csproj
```diff
@@ -17,8 +17,4 @@
     <PackageReference Include="Microsoft.EntityFrameworkCore.Sqlite" Version="2.0.0" />
     <PackageReference Include="System.Text.Encoding.CodePages" Version="4.4.0" />
   </ItemGroup>
-
-  <ItemGroup>
-    <Folder Include="Migrations\" />
-  </ItemGroup>
```
timabell: Was removing this intentional?

src/ManageCourses.Api/Controllers/DataController.cs
```diff
@@ -172,6 +176,48 @@ private void ProcessPayload(Payload payload)
                     }
                 );
             }
+
```
timabell: As per my comment on the importer, I think this is likely to need to become a separate operation before long. Especially as it's currently combined with dropping the existing data. Happy for this to go in and be iterated on.

src/ManageCourses.Domain/Models/McOrganisation.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace GovUk.Education.ManageCourses.Domain.Models
+{
+    public class McOrganisation
```
timabell: It's not immediately obvious to me what the "Mc" prefix means. Could this be made clearer somehow? Or could it be removed if it's not necessary?

src/ManageCourses.Domain/Models/McUser.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace GovUk.Education.ManageCourses.Domain.Models
+{
+    public class McUser
```
timabell: :hamburger:

PR #13, opened 2018-06-18
Api authorisation

MartinBentley:


src/ManageCourses.Api/Controllers/DataController.cs
```diff
@@ -4,6 +4,7 @@
 using GovUk.Education.ManageCourses.Domain.DatabaseAccess;
 using GovUk.Education.ManageCourses.Domain.Models;
 using Microsoft.AspNetCore.Mvc;
+using Remotion.Linq.Clauses;
```
timabell: Required?

src/ManageCourses.Api/Controllers/DataController.cs
```diff
@@ -65,6 +62,21 @@ private void ResetDatabase()
             _context.Save();
         }
 
+        private IEnumerable<Course> GetCoursesForUser(string email)
+        {
+            var userOrganisationNctls = _context.McOrganisationUsers.Where(o => o.Email == email).Select(x => x.NctlId);   
+            var mappedUcasCodes = _context.ProviderMappers.Where(uo => userOrganisationNctls.Contains(uo.NctlId)).Select(x => x.UcasCode);
```
the-daniel-rothig: Worth tracing this to see if all of this is evaluated server-side, don't want to retrieve that full result set

PR #20, opened 2018-06-21
New id fields

MartinBentley:


src/ManageCourses.Api/Controllers/DataController.cs
```diff
@@ -43,7 +43,7 @@ public DataController(IManageCoursesDbContext context)
         /// Exports the data.
         /// </summary>
         /// <returns>The exported data</returns>
-        [Authorize]
+        //[Authorize]
```
the-daniel-rothig: I'm guessing that's not intentional?

MartinBentley: Good spot.

PR #42, opened 2018-07-06
use GOV.UK notify for internal notifications

the-daniel-rothig:
### Context

We want to send emails to becomingateacher@digital.education.gov.uk when an access request is logged.

Using SMTP is so 2014 - tihs uses GOV.UK notify

### Guidance to review

need to set the following client secrets:
email:template_id 
email:api_key
email:user

all shared on slack.

Note: if api_key isn't set, EmailService.ShouldBeAbleToSend() is false i.e. no email gets sent. Useful for dev and test scenarios.


src/ManageCourses.Api/Startup.cs
```diff
@@ -54,9 +54,9 @@ public void ConfigureServices(IServiceCollection services)
             services.AddScoped<IDataService, DataService>();
 
             services.AddSingleton<IEmailService>(provider => new EmailService(
-                Configuration["email:host"],
-                Configuration["email:user"],
-                Configuration["email:password"]
+                Configuration["email:api_key"],
```
defong: Can you add token references here

https://github.com/DFE-Digital/manage-courses-api/blob/master/src/ManageCourses.Api/appsettings.SecretManager_Example.json

Can you also remove 
https://github.com/DFE-Digital/manage-courses-api/blob/396b7848229dcac4e5ebea979c031a2a94389eaf/src/ManageCourses.Api/appsettings.json#L15

Can we add the needed values here
https://dfedigital.atlassian.net/wiki/spaces/BaT/pages/389349377/Manage+Courses+API

src/ManageCourses.Api/Data/EmailService.cs
```diff
@@ -1,43 +1,59 @@
...
-            {
-                Subject = subject,
-                Body = message
             };
-            _smtpClient.Send(mailMessage);
+
+            _notificationClient.SendEmail(_user,  _templateId, templateValues);
+        }
+
+        public bool ShouldBeAbleToSend()
```
defong: This should be a private at least right?
May be this class and related, warrants unit test vs writing debugging code

Consider using a logger

the-daniel-rothig: Made reason more explicit by implementing factory pattern

src/ManageCourses.Api/Startup.cs
the-daniel-rothig: Ah I keep forgetting where the examples live!! thanks :)

defong:
Commit has progress from bool check for dev debugging
To convoluted solution to resolve a true or false.

src/ManageCourses.Api/Data/NullAccessRequestEmailService.cs
```diff
@@ -0,0 +1,12 @@
+﻿using GovUk.Education.ManageCourses.Domain.Models;
+
+namespace GovUk.Education.ManageCourses.Api.Data
+{
+    public class NullAccessRequestEmailService : IAccessRequestEmailService
```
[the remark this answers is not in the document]
the-daniel-rothig: It's not for debugging, it's for config resolution: One config means you can (and should) send emails --> AccessRequestEmailService. The other config means you can't (and shouldn't). Factory pattern for strategy selection's pretty standard, not sure why it would be overkill?

src/ManageCourses.Api/Data/AccessRequestEmailService.cs
```diff
@@ -51,9 +51,5 @@ public void SendAccessRequestEmailToSupport(AccessRequest accessRequest, McUser
             _notificationClient.SendEmail(_user,  _templateId, templateValues);
```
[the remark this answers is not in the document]
the-daniel-rothig: AccessRequestEmailService is no longer covering the case of not being able to send. this is now done by NullAccessRequestEmailService

src/ManageCourses.Api/Data/AccessRequestEmailService.cs
```diff
@@ -12,18 +12,31 @@ namespace GovUk.Education.ManageCourses.Api.Data
...
-        private string _templateId;
+        private readonly string _templateId;
 
         // This class only supports the RFC-compliant port 587
         private const int SupportedSmtpPort = 587;
 
         public AccessRequestEmailService(string apiKey, string templateId, string user)
         {
-            _notificationClient = String.IsNullOrWhiteSpace(apiKey) ? null : new NotificationClient(apiKey);
+            if (String.IsNullOrWhiteSpace(apiKey))
```
defong: Lowercase s
Else we write java

PR #43, opened 2018-07-06
Dead code purge

timabell:
### Context
There were some classes & tables left over from previous designs, these are not currently in use.

Remove 3 unused entities and all related code.


defong:
@MartinBentley Can you double check, these seems to look fine to me.

PR #44, opened 2018-07-06
Don't fall over when the user's email isn't known

timabell:
Log the guid from DfE SignIn to make it easier to support

### Context
Getting errors on prod but can't tell which user it's for.

```
Application started. Press Ctrl+C to shut down.
fail: Microsoft.EntityFrameworkCore.Query[10100]
      An exception occurred in the database while iterating the results of a query for context type 'GovUk.Education.ManageCourses.Domain.DatabaseAccess.ManageCoursesDbContext'.
      System.InvalidOperationException: Sequence contains no elements
         at System.Linq.Enumerable.First[TSource](IEnumerable`1 source)
         at lambda_method(Closure , QueryContext )
         at Microsoft.EntityFrameworkCore.Query.Internal.QueryCompiler.<>c__DisplayClass17_1`1.<CompileQueryCore>b__0(QueryContext qc)
System.InvalidOperationException: Sequence contains no elements
   at System.Linq.Enumerable.First[TSource](IEnumerable`1 source)
   at lambda_method(Closure , QueryContext )
at Microsoft.EntityFrameworkCore.Query.Internal.QueryCompiler.<>c__DisplayClass17_1`1.<CompileQueryCore>b__0(QueryContext qc)
```

### Changes proposed in this pull request

Handle unknown users more gracefully

* return 401 from API instead of crashing
* log guid from sign-in service to help troubleshooting

This makes the API more supportable.

When https://github.com/DFE-Digital/manage-courses-ui/pull/68 is merged as well the users will get a better experience.

PR #45, opened 2018-07-06
slightly better logging in the event of an auth error

the-daniel-rothig:
### Context
Some logging requested by PO

when a user arrives at our service but seems unauthenticated, this is getting logged now. Also more robust comparison.

### Guidance to review
n/a


benilovj:
I think #44 is a different take on this that fixes the exception and logs the GUID instead of email, reducing potential PII leakage.

timabell:
I prefer #44 but then I'm biased. I think #44 has a more complete implementation.

PR #47, opened 2018-07-09
Added the necessaries to make swagger fulfil api requirements

defong:
### Context
Making swagger useful

Added the ability for swagger to make authorised calls to the endpoints.

### Guidance to review

Our authorise endpoints require the header to contain

`Authorization` : `bearer ` + `{code}`

auth


timabell:
It would be good to give a hint to the user as to how one might obtain a bearer token.

defong:
A hint is great but 
* that's more of a blog post
* its more related to the external provider to assert  
* we just the consumer not the provider of the token

PR #48, opened 2018-07-09
Ignore case of email address from sign-in service

timabell:
### Context

Users registered with a different case of email address were not able to see their courses.

Make comparisons case-insensitive.


fofr:
Should there be a test for this?

timabell:
@fofr  I don't have strong feelings on the matter in this case. I'd be inclined to ship the fix as is for now I think.

defong:
I think this does warrant tests, simply it is a dead end situation for a user from step 1.

fofr:
A test would document the behaviour we expect and prevent future regressions.

timabell:
Okay then :-)

timabell:
Shape of services is wrong in the PR, abandoning and starting a new one.

PR #49, opened 2018-07-10
Fix zero courses for manchester

MartinBentley:


timabell:
I'd probably squash the two commits into one so that the tests stay green throughout.

PR #51, opened 2018-07-10
Managing multiple organisations

MartinBentley:
### Context

This is a breaking change, the matching UI PR is https://github.com/DFE-Digital/manage-courses-ui/pull/72


src/ManageCourses.Api/Data/DataService.cs
```diff
@@ -247,6 +249,16 @@ public OrganisationCourses GetCoursesForUser(string email)
 
             return returnCourses;
         }
+        /// <summary>
+        /// Gets a list of organisation the user in linked to
+        /// </summary>
+        /// <param name="email"></param>
+        /// <returns></returns>
```
defong: If you doing xml doc can you fill it in properly please

src/ManageCourses.Api/Data/DataService.cs
```diff
@@ -247,6 +249,11 @@ public OrganisationCourses GetCoursesForUser(string email)
 
             return returnCourses;
         }
+        public IEnumerable<UserOrganisation> GetOrganisationsForUser(string email)
```
defong: Is there a reason for this to be public?

defong:
Break out the test data in teest class to stand on its own.
If the class test is still too big
Break the test method in standalone class as well

timabell:
I think this might be returning the wrong orgIds, see comment on UI PR

PR #53, opened 2018-07-11
Userlog now persists even when we import new users

defong:
### Context
Similar to AccessRequest #50
Often we import new data, be it a new data dump from UCAS or a refined user list. This involves deleting old data, but the presence of access requests lead to FK violations.

This makes references in an UserLog to an existing user optional, while also adding UserEmail field. When the user list gets refreshed, existing requests are orphaned from their McUser but remain actionable since UserEmail can be matched up to the new corresponding McUser record

### DIFF
--- a/src/ManageCourses.Api/Services/UserLogService.cs
+++ b/src/ManageCourses.Api/Services/UserLogService.cs
@@ -15,18 +15,15 @@ namespace GovUk.Education.ManageCourses.Api.Services
             _context = context;
         }
 
-        public bool CreateOrUpdateUserLog(string signInUserId, string email) 
+        public bool CreateOrUpdateUserLog(string signInUserId, McUser user) 
         {
             var result = false;
             using (var transaction = ((DbContext)_context).Database.BeginTransaction()) 
             {
                 try
                 {
-                    var user = _context.McUsers.ByEmail(email).SingleOrDefault();
-
                     var userLog = _context.UserLogs
-                        .Include(x => x.User)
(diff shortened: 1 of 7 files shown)


src/ManageCourses.Api/Services/UserLogService.cs
```diff
@@ -26,7 +26,7 @@ public bool CreateOrUpdateUserLog(string signInUserId, string email)
 
                     var userLog = _context.UserLogs
                         .Include(x => x.User)
-                        .SingleOrDefault(x => (user != null ? x.User == user : true) && x.SignInUserId == signInUserId);
+                        .SingleOrDefault(x => (user != null ? x.User == user : true || !string.IsNullOrEmpty(x.UserEmail) ? string.Equals(x.UserEmail, email, StringComparison.InvariantCultureIgnoreCase) : true ) && x.SignInUserId == signInUserId);
```
the-daniel-rothig: way too much nested logic: two ternary operators, a string equality comparison and two boolean operators! This should be made more readable

the-daniel-rothig: I don't like that we duplicate the string comparison logic - can this be a method or extension method on McUser? (also to be used by `McUserQueryableExtensions.ByEmail`) cc. @timabell

timabell:
I think it would be better to have the import routine do an upsert instead of drop/recreate rather than weakening the integrity constraints. Can we discuss further before this goes in?

I also have a minor concern about keeping stale copies of email addresses lying around, especially in this GDPR obsessed climate.

PR #56, opened 2018-07-11
Set UserLog.UserId and AccessRequest.RequesterId to null when the cor…

the-daniel-rothig:
Set UserLog.UserId and AccessRequest.RequesterId to null when the corresponding user is deleted

### Context
We sometimes reimport users, which orphans access requests and user logs

Deleting users currently fails because the delete behaviour is set to `restrict`, rather than set null

### Guidance to review
There are two more entries in the DbContext fluent API (at the bottom)

EF Core requires us to introduce additional navigation properties from McUser to the dependent records, which is not particularly clean but unfortunately neccessary in order to set the OnDelete behaviour :(

### DIFF
--- a/src/ManageCourses.Domain/DatabaseAccess/ManageCoursesDbContext.cs
+++ b/src/ManageCourses.Domain/DatabaseAccess/ManageCoursesDbContext.cs
@@ -106,6 +106,18 @@ namespace GovUk.Education.ManageCourses.Domain.DatabaseAccess
                 .HasForeignKey(ucs => new { ucs.InstCode, ucs.CrseCode })
                 .HasPrincipalKey(cc => new { cc.InstCode, cc.CrseCode });
 
+            modelBuilder.Entity<AccessRequest>()
+                .HasOne(ar => ar.Requester)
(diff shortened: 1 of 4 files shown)


src/ManageCourses.Domain/DatabaseAccess/ManageCoursesDbContext.cs
```diff
@@ -106,6 +106,18 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
...
 
+            modelBuilder.Entity<AccessRequest>()
+                .HasOne(ar => ar.Requester)
+                .WithMany(u => u.AccessRequests)
+                .HasForeignKey(ar => ar.RequesterId)
+                .OnDelete(DeleteBehavior.SetNull);
+
+            modelBuilder.Entity<UserLog>()
+                .HasOne(ul => ul.User)
+                .WithMany(u => u.UserLogs)
```
the-daniel-rothig: Arguably the way it's written I think normally a McUser doesn't ever have more than 1 UserLog associated with them, but One-to-One relationships are even messier when it comes to delete behaviours :/

PR #57, opened 2018-07-12
Allow overriding auth for local development

timabell:
### Context
We can't fake DfE SignIn info for local development, but sometimes we want to impersonate a user locally. To reproduce an issue or try out a different set of data.

Add config and code that allows telling the API to not require DfE sign-in and instead just use a hard-coded email address from user secrets.


src/ManageCourses.Api/Middleware/DebugLoginHandler.cs
```diff
@@ -0,0 +1,41 @@
...
+
+            var identity = new ClaimsIdentity(
+                new[] {
+                            new Claim (ClaimTypes.NameIdentifier, "email"),
+                            new Claim (ClaimTypes.Email, Options.Email)
+                }, AuthenticationScheme, ClaimTypes.Email, null);
+
+            var princical = new ClaimsPrincipal(identity);
+
+            var ticket = new AuthenticationTicket(princical, BearerTokenDefaults.AuthenticationScheme);
```
defong: Wrong scheme

PR #59, opened 2018-07-16
Welcome email

defong:
### Context
Sending welcome email

When the user is logged for the first time send a welcome email

### Guidance to review
https://dfedigital.atlassian.net/wiki/spaces/BaT/pages/389349377/Manage+Courses+API

you need the following
        `api_key`
        `welcome_template_id`

And a working email address to receive the email.

omitting the `api_key` will turn off the sending of email

### DIFF
--- a/src/ManageCourses.Api/Services/UserLogService.cs
+++ b/src/ManageCourses.Api/Services/UserLogService.cs
@@ -28,23 +28,25 @@ namespace GovUk.Education.ManageCourses.Api.Services
                         .SingleOrDefault(x => x.SignInUserId == signInUserId);
 
                     var add = userLog == null;
+                    var now = DateTime.UtcNow;
                     if (add)
                     {
                         userLog = new UserLog
                         {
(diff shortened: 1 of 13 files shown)


src/ManageCourses.Api/Services/UserLogService.cs
```diff
@@ -26,21 +28,24 @@ public bool CreateOrUpdateUserLog(string signInUserId, McUser user)
...
                     }
 
                     userLog.User = user;
                     userLog.UserEmail = user.Email;
-                    userLog.LastLoginDateUtc = DateTime.UtcNow;
+                    userLog.LastLoginDateUtc = now;
 
                     if (add)
                     {
+                        _welcomeEmailService.Send(user);
```
timabell: It strikes me that the UserLog might be conceptually the wrong thing to be orchestrating the sending of the email. Think it breaks https://en.wikipedia.org/wiki/SOLID

PR #62, opened 2018-07-18
Script for pg database user, more readme info

timabell:
### Context

Things I needed to do to get this to run on my linux box

* Script for pg setup.
* Readme info on setting up user secrets in command line

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,9 +38,9 @@ An example of the config keys that are required for Secret Manager are available
 E.g.
 
     dotnet user-secrets set MANAGE_COURSES_POSTGRESQL_SERVICE_HOST localhost
-    dotnet user-secrets set PG_DATABASE manage
-    dotnet user-secrets set PG_USERNAME manageusr
-    dotnet user-secrets set PG_PASSWORD manageusr
+    dotnet user-secrets set PG_DATABASE the-database (will be created if missing and sufficient rights, e.g. 'manage')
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -25,12 +25,24 @@ In a windows command prompt:
...
 An example of the config keys that are required for Secret Manager are available from:
 
 	src\ManageCourses.Api\appsettings.SecretManager_Example.json
 
+E.g.
+
+    dotnet user-secrets set MANAGE_COURSES_POSTGRESQL_SERVICE_HOST localhost
+    dotnet user-secrets set PG_DATABASE manage
+    dotnet user-secrets set PG_USERNAME manageusr
+    dotnet user-secrets set PG_PASSWORD manageusr
```
defong: It is generally a bad practice to state a `password`  and other values as that become a given default to use.

I suggest you change it 
dotnet user-secrets set PG_PASSWORD `Your_password`

setup-pg-user.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+usr=manageusr
```
defong: Shouldn't you be using environment variables? or pass in the values?
script file name says `setup-pg-user.sh` shouldnt it be more generic to match up with the readme otherwise i would expect the script to setup pg user

timabell:
Updated to strip out passwords, could you re-review please @defong 
Thanks

timabell:
rebased

PR #68, opened 2018-07-20
Api key functionality

defong:
### Context
Api key based authentication.

Implement the Api key based authentication.

### Guidance to review
Refer to readme for config.

Basically the same as the other authenication mechisms but using api key instead.


src/ManageCourses.Api/Middleware/BearerTokenApiKeyHandler.cs
```diff
@@ -0,0 +1,57 @@
...
+                    }, BearerTokenApiKeyDefaults.AuthenticationScheme);
+
+                    var princical = new ClaimsPrincipal(identity);
+
+                    var ticket = new AuthenticationTicket(princical, BearerTokenApiKeyDefaults.AuthenticationScheme);
+
+                    return AuthenticateResult.Success(ticket);
+                }
+                else {
+                    return AuthenticateResult.Fail($"Invalid api key: {accessToken}");
```
timabell: I'm guessing we probably shouldn't include the key in the message in case it leaks the secret out to somewhere?

src/ManageCourses.Api/Middleware/BearerTokenApiKeyHandler.cs
```diff
@@ -0,0 +1,57 @@
...
+
+            return AuthenticateResult.NoResult();
+        }
+
+        protected override Task HandleChallengeAsync(AuthenticationProperties properties)
+        {
+            var authResult = this.HandleAuthenticateOnceSafeAsync().Result;
+            if (!authResult.Succeeded && authResult.Failure != null)
+            {
+                Logger.LogDebug(authResult.Failure, "Failed challenge");
```
timabell: Why do you log rather than throwing an exception here? If I understand the code correctly this isn't a normal execution path for either a client with the correct or incorrect api key.

src/ManageCourses.Api/Middleware/BearerTokenApiKeyOptions.cs
```diff
@@ -0,0 +1,21 @@
...
+namespace GovUk.Education.ManageCourses.Api.Middleware
+{
+    public class BearerTokenApiKeyOptions : AuthenticationSchemeOptions    
+    {
+        public string ApiKey { get; set; }
+
+        public override void Validate()
+        {
+            if (string.IsNullOrWhiteSpace(this.ApiKey)) {
+                throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, "The '{0}' option must be provided.", nameof(this.ApiKey)), nameof(this.ApiKey));
```
timabell: Is the culture info necessary?

src/ManageCourses.Api/Middleware/HttpRequestExtensions.cs
```diff
@@ -0,0 +1,19 @@
...
+    public static class HttpRequestExtensions
+    {
+        public static string GetAccessToken(this HttpRequest request)
+        {
+            var authorization = "Authorization";
+            var authorizationHeaderValues = request.Headers.ContainsKey(authorization) ?
+                ((string)request.Headers[authorization]).Split(' ') : new[] { "", "" };
+
+            var accessToken = authorizationHeaderValues.Length == 2 && authorizationHeaderValues[0].ToLowerInvariant().Equals("bearer") ?
+                authorizationHeaderValues[1] : "";
```
timabell: I think this would be hard to troubleshoot if any of the assumptions about the contents of the header failed.
I would throw if the assumptions are incorrect.

tests/ManageCourses.Tests/SmokeTests/SystemUnderTest.cs
```diff
@@ -22,14 +22,13 @@
 
 namespace GovUk.Education.ManageCourses.Tests.SmokeTests
 {
-
     [TestFixture]
-    public class DataServiceExportTests : ApiSmokeTestBase
+    [Category("Smoke")]
+    [Explicit]
+    public class SystemUnderTest : ApiSmokeTestBase
```
timabell: I'm not keen on the `SystemUnderTest` class name. I think it's too obscure. What was wrong with the original name? I thought it was more descriptive.

tests/ManageCourses.Tests/SmokeTests/SystemUnderTest.cs
```diff
@@ -66,6 +65,71 @@ public void DataExportWithEmptyCampus()
             Assert.AreEqual("Main campus site", campus.Name, "ProviderCourses.CourseDetails.Variants.Campuses.Name should be retrieved");
         }
 
+        [Test]
+        public void DataExport_badAccesCode_404()
```
timabell: I think the tests for testing the api key auth should live in a different test class even though they test the same endpoint.

src/ManageCourses.Api/Middleware/BearerTokenApiKeyHandler.cs
defong: That is the user input, you want it gone?

tests/ManageCourses.Tests/SmokeTests/SystemUnderTest.cs
defong: You don't really want to have to spin up a server for each test or test section. 

Consider will one endpoint vs another endpoint require a different server setup?

tests/ManageCourses.Tests/SmokeTests/SystemUnderTest.cs
defong: hmm partial class?


Reasons stills stand for system under test
* It is testing the exact same endpoint twice using the exact same configuration variated by its input
* both input are valid for their respective authentication mechanism.
* system is configured for only one of the authentication mechanism.

src/ManageCourses.Api/Middleware/HttpRequestExtensions.cs
defong: There is no assumption.

Headers was checked to see if it contains key auth.
auth value is split to array or a default of array with 2 empty string
if array is 2 return second value and if the first equals `bearer`

negative results will always be empty string 

throwing exception is expensive

if it contained auth  key but a null value then that is where the exception should be throw as that is beyond expectation

src/ManageCourses.Api/Middleware/BearerTokenApiKeyOptions.cs
defong: Origin
https://github.com/aspnet/Security/blob/442df16e265ab1955d7b8468cf1d8b38b8d915c7/src/Microsoft.AspNetCore.Authentication.OAuth/OAuthOptions.cs#L33

src/ManageCourses.Api/Middleware/BearerTokenApiKeyHandler.cs
defong: Exception are expensive and it needs to return expected behaviour to state the authentication result

ie a bad password is normal behaviour.

The log is there to catch exception and messages
ie
`return AuthenticateResult.Fail($"Invalid api key: {accessToken}")`

src/ManageCourses.Api/Middleware/BearerTokenApiKeyHandler.cs
defong: see below

src/ManageCourses.Api/Middleware/BearerTokenApiKeyHandler.cs
defong: Do you not want to know what values was attempted and failed?

timabell: Okay, I defer to your superior knowledge on the auth stuff.
I think the accessToken info should go, yes. If we find it hard to troubleshoot then we can think of another way of getting it.

src/ManageCourses.Api/Middleware/BearerTokenApiKeyOptions.cs
timabell: I'm not sure that answers my question other than implying it was a copy paste, but I don't mind if this goes in as is.

src/ManageCourses.Api/Middleware/HttpRequestExtensions.cs
timabell: The assumption I noticed was that the split would always result in two parts, and silently fail if that was not the case. Maybe that's never going to change but looking at the method in isolation it struck me as a smell that it was just swallowed. I wouldn't reject the pr for this though.

tests/ManageCourses.Tests/SmokeTests/SystemUnderTest.cs
timabell: Is there no way we can keep the same server while still having well factored code?

tests/ManageCourses.Tests/SmokeTests/SystemUnderTest.cs
timabell: see above comment thread.

tests/ManageCourses.Tests/SmokeTests/SystemUnderTest.cs
defong: Beyond immediate scope?

defong: This file in its entire is ugly.

src/ManageCourses.Api/Middleware/BearerTokenApiKeyHandler.cs
defong: That's why it's in debug, it's a value that was not expected.

timabell:
I would suggest splitting the test class into multiple classes, but...
Doing so would mean the expensive `OneTimeSetup` is run more times, and the solution to that is worse than the problem: https://stackoverflow.com/questions/23296454/how-can-i-execute-code-at-the-beginning-of-an-nunit-test-session-and-not-a-fixt

So let's merge this in and work out how to fix this, maybe a different test framework is required.

PR #71, opened 2018-07-23
Create delta import

MartinBentley:
### Context
At present the data import deletes all data before then re-inserting
This change implements a delta import for the users only
### Guidance to review
Initially this change was to be for all data tables so I created a new data helper class to deal with the various data entities.  This is probably over-kill for just one table, however I have decded to keep this new helper table as we will probably find it useful if we refactor the code to do this delta import on all tables at some later date.


src/ManageCourses.Api/Data/DataService.cs
```diff
@@ -17,10 +19,20 @@ public DataService(IManageCoursesDbContext context)
         }
 
         #region Import
+        /// <summary>
+        /// Processes data in the payload object into the database as an upsert/delta
+        /// </summary>
+        /// <param name="payload"></param>
```
defong: Fill it in, this gets shown in swagger.

src/ManageCourses.Api/Data/DataService.cs
```diff
@@ -17,10 +19,20 @@ public DataService(IManageCoursesDbContext context)
...
         public void ProcessPayload(Payload payload)
         {
             ResetDatabase();
 
+            var userDatahelper = new UserDataHelper(_context, payload.Users.Where(u => !string.IsNullOrWhiteSpace(u.Email)).ToList());
```
defong: Why? Move it to a service.

src/ManageCourses.Api/Data/UserDataHelper.cs
```diff
@@ -0,0 +1,104 @@
...
+        private readonly IManageCoursesDbContext _context;
+        private readonly IReadOnlyCollection<McUser> _dataList;
+        private ImportMapper<McUser> _mapper;
+        /// <summary>
+        /// constructor for the class. Initialises the dataset and context.
+        /// Generates the diff (additions, updates, deletions)
+        /// </summary>
+        /// <param name="context">context for the db</param>
+        /// <param name="dataList">new dataset coming from the importer</param>
+        public UserDataHelper(IManageCoursesDbContext context, IReadOnlyCollection<McUser> dataList)
```
defong: Change this into a service there is no reason not to.

defong: Also inject a logger, so we can have some perception of whats going on.

src/ManageCourses.Api/Data/UserDataHelper.cs
```diff
@@ -0,0 +1,104 @@
...
+            {
+                _context.McUsers.AddRange(_mapper.Additions);
+                _context.McUsers.RemoveRange(_mapper.Deletes);
+                DoUpdates();
+                _context.Save();
+                returnBool = true;
+            }
+            catch (Exception e)
+            {
+                //write to error log?
```
defong: Transaction maybe?

src/ManageCourses.Api/Data/UserDataHelper.cs
defong: Unit test?

MartinBentley: All done

src/ManageCourses.Api/Data/UserDataHelper.cs
MartinBentley: An UpsertResult class has been added that holds an error message or upsert details if successful

PR #81, opened 2018-08-01
Reset smoke data

timabell:
### Context

Smoke test `ApiEndpointTests.DataExport_WithEmptyCampus` would fail if run individually after other test runs because there was no cleanup of the data.

This PR adds cleanup before & after each smoke test run, fixing the flapping test.

* Remove hard-coded username/password etc, rely on single user-secrets for all db config (test + captured host)
* Code cleanup
* New readme for context on tests

### DIFF
--- a/src/ManageCourses.Api/Startup.cs
+++ b/src/ManageCourses.Api/Startup.cs
@@ -29,12 +29,11 @@ namespace GovUk.Education.ManageCourses.Api
         // This method gets called by the runtime. Use this method to add services to the container.
         public void ConfigureServices(IServiceCollection services)
         {
-            var connectionString = GetConnectionString(Configuration);
-
             services.AddEntityFrameworkNpgsql()
                 .AddDbContext<ManageCoursesDbContext>(
                     options => options.UseNpgsql(
-                        connectionString, b => b.MigrationsAssembly((typeof(ManageCoursesDbContext).Assembly).ToString())
+                        GetConnectionString(Configuration),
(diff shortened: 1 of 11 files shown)


src/ManageCourses.Api/Startup.cs
```diff
@@ -113,7 +114,16 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env, ManageCo
...
-        public static string GetConnectionString(IConfiguration config)
+        /// <summary>
+        /// Build a postgres connection string from configuration data
+        /// </summary>
+        /// <param name="config"></param>
+        /// <param name="dbNameSuffix">
+        ///     Optional. String to append to the database name, e.g. "test" to use "manage-test" as the database name.
+        ///     Intended to be used for building various test databases
+        /// </param>
+        /// <returns></returns>
```
defong: Fill it in...

timabell:
@defong I agree we could use a broader improvement. Added https://trello.com/c/mgvkSKxG to our board.

timabell:
@defong  re `dbNameSuffix` - we have two sets of integration tests:

1) "smoke" (with a real db and a real http endpoint) and 
2) "DbIntegration" with service classes tested against a real db

They both generate a real postgres db. I've consolidated the code for configuring what database/username/password etc is used to the standard user secrets / configuration stuff with standard config names, but wanted to keep having two different databases. This is a slightly hacky way of achieving that. A bigger piece of work might be to split the tests into 3x projects and rethink the sharing between the two integration tests and the api itself.

PR #85, opened 2018-08-03
Return 422 if invited user doesn't exist

timabell:
### Context

API was returning 500 server error for unknown users, which is indistinguishable from any other failure.


timabell:
[Invite - Return something other than 500 status for unknown user](https://trello.com/c/L5aC5wxP/70-invite-return-something-other-than-500-status-for-unknown-user)

benilovj:
Should this be tested?

timabell:
Good point. Yes.

timabell:
Changed error http status code in use to 400 based on Sau's PR questions, and discovering the spec had changed. Left [some notes on the relevant stackoverflow questions](https://stackoverflow.com/questions/3050518/what-http-status-response-code-should-i-use-if-the-request-is-missing-a-required/10323055#comment90367972_3050518) to help anyone else avoid the outdated answers that tripped me up.

PR #92, opened 2018-08-13
Enrichment endpoint stubs

timabell:
### Context

To allow UI work to continue while we build a real implementation here's a PR that adds stub endpoint

* POST swallow any data and does notthing with it
* GET returns an empty initialized structure

Paired with @MartinBentley 

Add a stub load/save enrichment data endpoint, with smoke test.

---

Here's the GET in swagger (real + example json response):

vm_300


timabell:
[Publish info 'About you org'](https://trello.com/c/KbZIyaqd/108-publish-info-about-you-org)

PR #96, opened 2018-08-15
Institution-wise import

defong:
### Context
The basic flow is wrapping it in a transaction for deleting and adding and also for recovery.

When a transaction completes it either commits or rollback if it fails.

It has a recovery for the purpose of deleting the item concerned and adding a new version back if it fails to add the new version it uses the deleted copy as recovery as ef do insert before deletion

Isolate failures in UCAS importer for robustness

### DIFF
--- a/src/ManageCourses.Api/Services/Data/DataService.cs
+++ b/src/ManageCourses.Api/Services/Data/DataService.cs
@@ -34,6 +34,7 @@ namespace GovUk.Education.ManageCourses.Api.Services.Data
         {
             var allInstitutions = _context.UcasInstitutions.ToList();
             
+            _logger.LogWarning("Beginning UCAS import");
             _logger.LogInformation($"Upserting {allInstitutions.Count()} institutions");
             int processed = 0;
             foreach (var inst in allInstitutions)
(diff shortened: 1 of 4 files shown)


timabell:
Chatted about this with @defong 
We need a high level of confidence in this code as it'll be painful if this fails in prod. Since this was branched off before the enrichment was in place it needs to be validated against the current enrichment code / data. I suggest we extend the DbIntegration tests to do an enrichment and then an import to check FKs behave and data is not lost.

timabell:
Rebased

src/ManageCourses.Api/Services/Data/DataService.cs
```diff
@@ -30,18 +32,124 @@ public DataService(IManageCoursesDbContext context, IDataHelper dataHelper, ILog
...
-            ResetUcasSchema();
+            var allInstitutions = _context.UcasInstitutions.ToList();
+            
+            _logger.LogWarning("Beginning UCAS import");
+            _logger.LogInformation($"Upserting {allInstitutions.Count()} institutions");
+            int processed = 0;
+            foreach (var inst in allInstitutions)
+            {
+                var deleted = AttemptTransaction(inst.InstCode, "delete for replace", () => DeleteForInstitution(inst.InstCode));
+                var added = AttemptTransaction(inst.InstCode, "insert", () => AddForInstitution(inst.InstCode, payload));
```
timabell: Why wouldn't we just wrap the delete+add in a single transaction and do away with all the EF state tracking and attempted rollbacks?

timabell: Answer: was already attempted.

timabell:
Additional test coverage for interaction with enrichment provided in #100

PR #98, opened 2018-08-15
Add FromBody attribute to enrichment

timabell:
Without this the post from the UI isn't read properly


benilovj:
Maybe worth beefing up the asserts on the smoke test...

PR #99, opened 2018-08-15
Changes to Enrichment service and tests

MartinBentley:
### Context
The Api wasn't working because of lowercase inst codes
This has now been rectified

### DIFF
--- a/src/ManageCourses.Api/Services/UserInstitution.cs
+++ b/src/ManageCourses.Api/Services/UserInstitution.cs
@@ -1,0 +1,14 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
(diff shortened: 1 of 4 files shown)


src/ManageCourses.Api/Services/UserInstitution.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using GovUk.Education.ManageCourses.Domain.Models;
+
+namespace GovUk.Education.ManageCourses.Api.Services
+{
+    public class UserInstitution
```
timabell: This class smells a bit funny but I haven't got any suggestions off the top of my head so :+1:

PR #100, opened 2018-08-16
Remove FK relationship btw Enrichment and Institution

the-daniel-rothig:
### Context

UCAS deletes-and-recreates lots of institutions. It will fail if there are enrichments as FK violations protect from deletion

Remove FK relationships

### Guidance to review

See the test, and the migration script

### DIFF
--- a/src/ManageCourses.Domain/DatabaseAccess/ManageCoursesDbContext.cs
+++ b/src/ManageCourses.Domain/DatabaseAccess/ManageCoursesDbContext.cs
@@ -115,10 +115,7 @@ namespace GovUk.Education.ManageCourses.Domain.DatabaseAccess
                 .OnDelete(DeleteBehavior.SetNull);
 
             modelBuilder.Entity<InstitutionEnrichment>()
-                .HasOne(ie => ie.UcasInstitution)
-                .WithMany(ui => ui.InstitutionEnrichments)
-                .HasForeignKey(ie => ie.InstCode)
-                .HasPrincipalKey(ui => ui.InstCode);
+                .HasIndex(x => x.InstCode);
 
             base.OnModelCreating(modelBuilder);
         }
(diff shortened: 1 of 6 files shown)


the-daniel-rothig:
[Remove Foreign Key relationships between Enrichment data and ucas data](https://trello.com/c/eI30OB1P/143-remove-foreign-key-relationships-between-enrichment-data-and-ucas-data)

MartinBentley:
Why are we doing this? Why don't we just remove all the dependencies before removing the institution.
If we remove the FK's then delete the institution then we end up with redundant data

timabell:
Could we disable the FK temporarily while the import runs instead?

https://stackoverflow.com/questions/38112379/disable-postgresql-foreign-key-checks-for-migrations/38113838#38113838

timabell:
Test failure:

> Failed   EnrichmentDataSurvivesDeleteAndRecreate
> Error Message:
>  System.NullReferenceException : Object reference not set to an instance of an object.
> Stack Trace:
>    at GovUk.Education.ManageCourses.Api.Services.EnrichmentService.Convert(InstitutionEnrichment source) in /home/travis/build/DFE-Digital/manage-courses-api/src/ManageCourses.Api/Services/EnrichmentService.cs:line 153
>    at GovUk.Education.ManageCourses.Api.Services.EnrichmentService.GetInstitutionEnrichment(String instCode, String email) in /home/travis/build/DFE-Digital/manage-courses-api/src/ManageCourses.Api/Services/EnrichmentService.cs:line 35
>    at GovUk.Education.ManageCourses.Tests.DbIntegration.EnrichmentServiceTests.EnrichmentDataSurvivesDeleteAndRecreate() in /home/travis/build/DFE-Digital/manage-courses-api/tests/ManageCourses.Tests/DbIntegration/EnrichmentServiceTests.cs:line 324

timabell:
We don't need this any more. The importer uses non-destructive upsert for institution now.

timabell:
Was going to kill this PR as the import now works with this FK in place (due to change to upsert of UcasInstitutions in master), however on discussion with @the-daniel-rothig  it has been agreed we should maintain an architectural separation between Ucas data and Enrichment data, so the FK should not exist after all.
Work on test coverage has been reworked and this PR been force pushed to show the separation of different pieces of work.

PR #101, opened 2018-08-17
Publish organisation

MartinBentley:
### Context
This is a change to the Api that sets an enrichment records to status="Published"
New columns in the InstitutionEnrichment table (status, LastPublishedDate)
### Guidance to review
Here is the workflow:
1) When an enrichment is initially saved the process creates a new record with a status of "Draft".
2) When the record is saved again the process updates the record (no new record)
3) When the record is published the status is set to Published.
 4) When the enrichment is saved again, a new record is created with th LastPublishedDate set to the LastPublishedDate of the previous Published record.


src/ManageCourses.Api/Model/UcasInstitutionEnrichmentGetModel.cs
```diff
@@ -9,8 +10,11 @@ public UcasInstitutionEnrichmentGetModel()
             EnrichmentModel = new InstitutionEnrichmentModel();
         }
         public InstitutionEnrichmentModel EnrichmentModel { get; set; }
-
         public DateTime? CreatedTimestampUtc { get; set; }
         public DateTime? UpdatedTimestampUtc { get; set; }
+        public DateTime? LastPublishedTimestampUtc { get; set; }
```
timabell: We can infer last published with `.Where(x => x.Status == published).OrderBy(x => x.Updated).First()`, do we need to duplicate this data?

src/ManageCourses.Api/Services/EnrichmentService.cs
```diff
@@ -21,22 +21,94 @@ public EnrichmentService(IManageCoursesDbContext context)
...
+            }
+            _context.Save();
+        }
+        /// <summary>
+        /// Changes the status of the latest draft record to published
+        /// </summary>
+        /// <param name="instCode">institution code of the enrichemtn to be published</param>
+        /// <param name="email">email of the user</param>
+        /// <returns>true if successful</returns>
+        public bool PublishInstitutionEnrichment(string instCode, string email)
```
timabell: Use exceptions not return values to indicate error conditions.

Reject

tests/ManageCourses.Tests/DbIntegration/EnrichmentServiceTests.cs
```diff
@@ -157,6 +163,20 @@ public void Test_InstitutionEnrichment_workflow_should_not_error()
...
+            var publishRecord = enrichmentService.GetInstitutionEnrichment(_providerInstCode.ToLower(), _email);
+            publishRecord.Status.Should().BeEquivalentTo(EnumStatus.Published);
+            publishRecord.LastPublishedTimestampUtc.Should().NotBeNull();
+            //test save again after publish
+            enrichmentService.SaveInstitutionEnrichment(model, _providerInstCode.ToLower(), _email);
+            var updateResult2 = enrichmentService.GetInstitutionEnrichment(_providerInstCode, _email);
+            updateResult2.EnrichmentModel.TrainWithDisability.Should().BeEquivalentTo(trainWithDisabilityText);
+            updateResult2.EnrichmentModel.TrainWithUs.Should().BeEquivalentTo(trainWithUsText);
+            updateResult2.Status.Should().BeEquivalentTo(EnumStatus.Draft);
+            updateResult2.LastPublishedTimestampUtc.ToString().Should().BeEquivalentTo(publishRecord.LastPublishedTimestampUtc.ToString());
```
timabell: Might be good to count the number of records of each type that ended up in the db at this point,

src/ManageCourses.Api/Model/UcasInstitutionEnrichmentGetModel.cs
MartinBentley: No we need to keep these separate. The last published date can be null for a new draft record even though the updated timestamp is set on each subsequent. Also a new draft needs to have this field set only if a previous draft has been published for the institution.

src/ManageCourses.Api/Services/EnrichmentService.cs
MartinBentley: The bool will never indicate an error condition. It simply indicates whether or not the publish succeeded. For example if a draft enrichment record does not exist then this returns false when is quite reasonable and not an error. There is no try/catch here so all exceptions will just throw at that point.

PR #105, opened 2018-08-21
Nctl orgs

the-daniel-rothig:
### Context

When adding users, they usually know which "NCTL Organisation" they belong to, and what "UCAS institutions" they own, but our `org_id` concept is new and unknown. To action access requests, we need to be able to inspect which NCTL Organisation belong to which `org_id`

This introduces a new entity `NctlOrganisation`, with a many-to-one relationship to `McOrganisation`. The new entities are added to the ReferenceData Payload class and import method.

### Guidance to review

The following file imports the production reference data: [import.zip](https://github.com/DFE-Digital/manage-courses-api/files/2307527/import.zip)

### DIFF
--- a/src/ManageCourses.Api/Model/ReferenceDataPayload.cs
+++ b/src/ManageCourses.Api/Model/ReferenceDataPayload.cs
@@ -8,12 +8,14 @@ namespace GovUk.Education.ManageCourses.Api.Model
         public ReferenceDataPayload()
         {
             this.Organisations = new List<McOrganisation>();
+            this.NctlOrganisation = new List<NctlOrganisation>();
             this.Institutions = new List<UcasInstitution>();
             this.OrganisationInstitutions = new List<McOrganisationInstitution>();
             this.OrganisationUsers = new List<McOrganisationUser>();
             this.Users = new List<McUser>();
         }
         public IEnumerable<McOrganisation> Organisations { get; set; }
+        public IEnumerable<NctlOrganisation> NctlOrganisation { get; set; }
         public IEnumerable<UcasInstitution> Institutions { get; set; }
         public IEnumerable<McOrganisationInstitution> OrganisationInstitutions { get; set; }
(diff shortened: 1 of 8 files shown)


src/ManageCourses.Domain/Migrations/ManageCoursesDbContextModelSnapshot.cs
```diff
@@ -532,6 +555,15 @@ protected override void BuildModel(ModelBuilder modelBuilder)
                         .HasPrincipalKey("OrgId");
                 });
 
+            modelBuilder.Entity("GovUk.Education.ManageCourses.Domain.Models.NctlOrganisation", b =>
+                {
+                    b.HasOne("GovUk.Education.ManageCourses.Domain.Models.McOrganisation", "McOrganisation")
+                        .WithMany("NctlOrganisations")
+                        .HasForeignKey("OrgId")
+                        .HasPrincipalKey("OrgId")
+                        .OnDelete(DeleteBehavior.Cascade);
```
benilovj: I'm not convinced we should be deleting on cascade. As I understand it, an `NctlOrganisation` can exist without a `McOrganisation` and vice versa.

the-daniel-rothig:
[Update org mappings and merge accounts in our database](https://trello.com/c/xC4GcSNa/63-update-org-mappings-and-merge-accounts-in-our-database)

src/ManageCourses.Domain/Migrations/ManageCoursesDbContextModelSnapshot.cs
the-daniel-rothig: :-1: NctlOrganisation can't exist without McOrganisation, as it is metadata on McOrganisation, hence the cascade

benilovj: ok, we'll just need to ensure that we have a McOrganisation for every org with an allocation, whenever we update allocations info.

PR #109, opened 2018-08-23
Course enrichment domain

timabell:
### Context

Stub endpoint for course enrichment for coding against

3 new enrichment endpoints - course enrichment get, set, publish

They don't do anything yet.


src/ManageCourses.Api/Model/UcasCourseEnrichmentGetModel.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace GovUk.Education.ManageCourses.Api.Model
+{
+    public class UcasCourseEnrichmentGetModel
+    {
+        public UcasCourseEnrichmentGetModel()
+        {
+            EnrichmentModel = new CourseEnrichmentModel();
+        }
+        public CourseEnrichmentModel EnrichmentModel { get; set; }
```
the-daniel-rothig: Doesn't this need identifying informmation aka. a `courseCode` and a `providerCode`?

the-daniel-rothig:
Oh also, you probably want to bump the API version number to `0.6.1`

defong:
No @the-daniel-rothig MINOR version when you add functionality in a backwards-compatible manner

timabell:
@the-daniel-rothig re versions
I beg to differ
> MINOR version when you **add functionality** in a backwards-compatible manner, and
> PATCH version when you make backwards-compatible bug fixes.

the-daniel-rothig:
Oops I stand corrected, sry.

src/ManageCourses.Api/Model/UcasCourseEnrichmentGetModel.cs
timabell: They are in the url, captured as separate string parameters.
Just tried out some variations locally with the fields added to the model.

* Adding to model: now have duplicate places to post this info, bits of url are not bound to fields of model
* Adding to model and removing controller action params: results in broken url pattern (replacement bits disappear, losing our nice restful url setup)

PR #111, opened 2018-08-23
Status for course

MartinBentley:
### Context
This change is to surface the course status.
The UcasCourse table now has a Status. (other fields have been added but these are not require atm)
The importer now provides the payload with the course status.
### Guidance to review
Status should now be return from the Api as part of the Course data.
The School object will now hold the status as this is need for the UI


src/ManageCourses.Api/Services/Data/DataService.cs
```diff
@@ -426,7 +421,8 @@ private Course LoadCourse(IReadOnlyList<UcasCourse> courseRecords)
                 Address4 = courseRecord.UcasCampus.Addr4,
                 PostCode = courseRecord.UcasCampus.Postcode,
                 ApplicationsAcceptedFrom = courseRecord.CrseOpenDate,
-                Code = courseRecord.UcasCampus.CampusCode
+                Code = courseRecord.UcasCampus.CampusCode,
+                Status = courseRecord.Status
```
timabell: Add the trailing commas to lists to make diffs cleaner in future.

Minor.

tests/ManageCourses.Tests/DbIntegration/DataServiceTests.cs
```diff
@@ -254,6 +254,26 @@ public void GetCoursesShouldReturnData()
...
         }
+        [Test]
+        public void GetCoursesShouldReturnStatus()
+        {
+            const int numOrgs = 5;
+            const int numCourses = 6;
+            LoadData(TestUserEmail1, numOrgs, numCourses);
+            var orgList = DataService.GetOrganisationsForUser(TestUserEmail1).ToList();
+            Assert.IsTrue(orgList.Count == numOrgs);
+            Assert.IsTrue(orgList.All(c => c.TotalCourses == numCourses));
```
timabell: Use fluent assertions throughout for nicer syntax and better test failure messages.

Reject.

tests/ManageCourses.Tests/DbIntegration/DataServiceTests.cs
```diff
@@ -435,7 +455,8 @@ private static UcasPayload GetUcasCoursesPayload()
                     new UcasCourse
                     {
                         InstCode = InstCode1,
-                        CrseCode = "CourseCode_1"
+                        CrseCode = "CourseCode_1",
+                        Status = "N"
```
timabell: you can see in this the effect on the diff of not previously having the trailing comma, it turns what would have been a one-line addition into a two line addition and one line deletion

src/ManageCourses.Api/Services/Data/DataService.cs
MartinBentley: Done

tests/ManageCourses.Tests/DbIntegration/DataServiceTests.cs
MartinBentley: Done

tests/ManageCourses.Tests/DbIntegration/DataServiceTests.cs
MartinBentley: All done

PR #120, opened 2018-09-01
cache access tokens to reduce the number of calls to DfE Sign In

the-daniel-rothig:
### Context

https://trello.com/c/4H5Eukfb/176-performance-issues-on-dev-when-enriching-courses

Cache Access Tokens in the database. During authentication, we first check if the access token is already known to the system as an McSession.

To lower the risk of access token clashes, a life span of 30 minutes is enforced.

### Guidance to review

Theres two parts to this: The schema changes (introducing a new entity `McSession`), and the bearer token handler changes. The latter is more interesting so I'd recommend starting from `BearerTokenHandler.cs`

### DIFF
--- a/src/ManageCourses.Api/Services/Users/UserService.cs
+++ b/src/ManageCourses.Api/Services/Users/UserService.cs
@@ -47,14 +47,14 @@ namespace GovUk.Education.ManageCourses.Api.Services.Users
             LogLogin(mcUser);
             UpdateMcUserFromSignIn(mcUser, userDetails);
 
-            if (!_context.McSessions.Any(x => x.AccessToken == accessToken && x.CreatedUtc > DateTime.UtcNow.AddMinutes(-30)))
+            if (!_context.McSessions.Any(x => x.AccessToken == accessToken && x.CreatedUtc > _clock.UtcNow.AddMinutes(-30)))
             {
                 _context.McSessions.Add(new McSession{
                     AccessToken = accessToken,
                     McUser = mcUser,
                     Email = mcUser.Email,
                     Subject = userDetails.Subject,
-                    CreatedUtc = DateTime.UtcNow
+                    CreatedUtc = _clock.UtcNow
                 });
             }
 
--- a/src/ManageCourses.Api/Services/Users/UserService.cs
+++ b/src/ManageCourses.Api/Services/Users/UserService.cs
@@ -52,7 +52,6 @@ namespace GovUk.Education.ManageCourses.Api.Services.Users
                 _context.McSessions.Add(new McSession{
                     AccessToken = accessToken,
                     McUser = mcUser,
-                    Email = mcUser.Email,
                     Subject = userDetails.Subject,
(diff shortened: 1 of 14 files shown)


the-daniel-rothig:
[Performance issues on dev when enriching courses](https://trello.com/c/4H5Eukfb/176-performance-issues-on-dev-when-enriching-courses)

defong:
It probably just better to scrap dfe sign in dependency, and actually just have a 2 way communication rather then 4 way communication vector that we will now have.

src/ManageCourses.Domain/DatabaseAccess/ManageCoursesDbContext.cs
```diff
@@ -123,6 +123,17 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
 
             modelBuilder.Entity<InstitutionEnrichment>()
                 .HasIndex(x => x.InstCode);
+                
+            modelBuilder.Entity<McSession>()
+                .HasOne(x => x.McUser)
+                .WithMany(u => u.Sessions)
+                .HasForeignKey(x => x.Email)
+                .HasPrincipalKey(u => u.Email)
```
timabell: can we FK on id now? This is another one we'll have to undo

src/ManageCourses.Api/Services/Users/UserService.cs
```diff
@@ -44,6 +46,18 @@ public async Task UserSignedInAsync(JsonUserDetails userDetails)
             }
             LogLogin(mcUser);
             UpdateMcUserFromSignIn(mcUser, userDetails);
+
+            if (!_context.McSessions.Any(x => x.AccessToken == accessToken && x.CreatedUtc > _clock.UtcNow.AddMinutes(-30)))
```
timabell: should we make the timeout configurable?

src/ManageCourses.Api/Startup.cs
```diff
@@ -92,7 +92,7 @@ public void ConfigureServices(IServiceCollection services)
         // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
         public void Configure(IApplicationBuilder app, IHostingEnvironment env, ManageCoursesDbContext dbContext)
         {
-            dbContext.Database.Migrate();
+            //dbContext.Database.Migrate();
```
timabell: why is this commented out?

src/ManageCourses.Domain/Migrations/ManageCoursesDbContextModelSnapshot.cs
```diff
@@ -242,15 +242,15 @@ protected override void BuildModel(ModelBuilder modelBuilder)
                     b.Property<DateTime>("CreatedUtc")
                         .HasColumnName("created_utc");
 
-                    b.Property<string>("Email")
-                        .HasColumnName("email");
+                    b.Property<int?>("McUserId")
```
timabell: I think you can get this to not be nullable by adding `[Required]` to the McUser property on the session model

PR #121, opened 2018-09-03
Add enrichment status to course

timabell:
### Context

We want to show the course enrichment status next to the courses listed in the UI.

This PR extends the data returned by the API for the list of courses to included the enrichment status.

Also fixed some performance issues by making some queries more EF friendly. Now assuming convention of storage for uppercase/lowercase in db keys to avoid need to do translation during querying.

Matching UI PR: https://github.com/DFE-Digital/manage-courses-ui/pull/142

* Extend retrieved ucas course with current enrichment status
* Migrate existing data to all upper case keys
* Fix EF queries by assuming uppercase instead of converting.
* Update existing tests to use uppercase convention to avoid a couple of failures.
* Extend existing test coverage to cover new fields.

There's a wip test for more coverage but will do this in another PR as this PR is holding up a bunch of other things.


src/ManageCourses.Api/Services/Data/DataService.cs
```diff
@@ -373,19 +375,22 @@ private InstitutionCourses LoadCourses(IReadOnlyList<UcasCourse> courseRecords)
...
-                    returnCourses.Courses.Add(LoadCourse(courseRecords.Where(c => c.CrseCode == courseCode).ToList()));
+                    returnCourses.Courses.Add(LoadCourse(courseRecords.Where(c => c.CrseCode == courseCode).ToList(), enrichmentMetadata));
                 }
             }
 
             return returnCourses;
 
         }
-        private Course LoadCourse(IReadOnlyList<UcasCourse> courseRecords)
+        private Course LoadCourse(IReadOnlyList<UcasCourse> courseRecords, IList<UcasCourseEnrichmentGetModel> enrichmentMetadata)
```
MartinBentley: Can we make this IEnumerable

src/ManageCourses.Api/Services/Data/DataService.cs
```diff
@@ -232,7 +237,7 @@ private UcasPayload AddForInstitution(string instCode, UcasPayload payload)
             }
             return payload;
         }
-        
+
         public UserOrganisation GetOrganisationForUser(string email, string instCode)
         {
             var org = GetUserOrganisation(email, instCode);
```
MartinBentley: Can we the validation method call here also

src/ManageCourses.Api/Model/UcasCourseEnrichmentGetModel.cs
```diff
@@ -16,5 +16,7 @@ public UcasCourseEnrichmentGetModel()
         public int CreatedByUserId { get; set; }//TODO should be a user object however it causes an error 
         public int UpdatedByUserId { get; set; }//TODO should be a user object however it causes an error 
         public EnumStatus Status { get; set; }
+        public string InstCode { get; set; }
+        public string CourseCode { get; set; }
     }
 }
```
MartinBentley: This is virtually identical to the UcasInstitutionEnrichmentGetModel. Can we create a base class for these types

timabell: card it up ;-)

PR #123, opened 2018-09-04
Ditch McUser ByEmail in favour of EF-friendly method

timabell:
### Context

ByEmail wasn't being converted into LINQ by EF

* Throw it away and use the newer FromSql method.
* Turn `DataServiceExportTests` into an integration test because the in-memory db used in the unit tests doesn't support `FromSql()`
* Do backflips to get the test data to be inserted only once and to obey FK and unique constraints in the real postgres database.


src/ManageCourses.Domain/Models/McOrganisationUser.cs
```diff
@@ -1,5 +1,8 @@
-﻿namespace GovUk.Education.ManageCourses.Domain.Models
+﻿using System.Diagnostics;
+
+namespace GovUk.Education.ManageCourses.Domain.Models
 {
+    [DebuggerDisplay("{DebuggerDisplay,nq}")]
```
defong: Should this be here?

tests/ManageCourses.Tests/DbIntegration/DataServiceExportTests.cs
```diff
@@ -1,47 +1,46 @@
...
     {
-        private ManageCoursesDbContext _dbContext;
         private DataService _dataService;
 
-        [OneTimeSetUp]
-        public void Setup()
+        protected override void Setup()
         {
-            _dbContext = TestHelper.GetFakeData(EnumTestType.DataService);            
+            new TestHelper(Context).BuildFakeDataForService();
```
defong: new it up and don't use it?

tests/ManageCourses.Tests/UnitTesting/Helpers/TestHelper.cs
```diff
@@ -1,314 +1,305 @@
...
-                case EnumTestType.DataHelper:
-                    BuildFakeDataForHelper();
-                    break;
-            }
-            return _dbContext;
+            DbContext = dbContext;
         }
 
-        private static void SetupContext()
+        private void SetupInMemoryContext()
```
defong: Is this actually getting used?

tests/ManageCourses.Tests/UnitTesting/Helpers/TestHelper.cs
```diff
@@ -1,314 +1,305 @@
...
         {
             var optionsBuilder = new DbContextOptionsBuilder<ManageCoursesDbContext>();
             optionsBuilder.UseInMemoryDatabase("dbInMemory");
-            _dbContext = new ManageCoursesDbContext(optionsBuilder.Options);
+            DbContext = new ManageCoursesDbContext(optionsBuilder.Options);
         }
 
-        private static void BuildFakeDataForHelper()
+        public void BuildFakeDataForHelper()
```
defong: Static and pass in context?

timabell:
Okay so it was EF friendly after all. I'm just shell-shocked from having used EF6 for so long.

Going to keep the change anyway as it's generally an improvement and removes duplicate way of doing same thing.

src/ManageCourses.Domain/Models/McOrganisationUser.cs
timabell: yes :-)

tests/ManageCourses.Tests/DbIntegration/DataServiceExportTests.cs
timabell: It's used immediately by calling BuildFakeDataForService

tests/ManageCourses.Tests/UnitTesting/Helpers/TestHelper.cs
timabell: yes

tests/ManageCourses.Tests/UnitTesting/Helpers/TestHelper.cs
timabell: lots of things in this class depend on hitting the context field, I think including methods this calls.

PR #125, opened 2018-09-04
Added field to CourseEnrichmentModel. bumped version. Do Not Merge

MartinBentley:
### Context
There is a new field to capture in the gui where a user selects "Other" as an option for Course Length.
The user should then enter text into the new field,


the-daniel-rothig:
I'm not sure this should be a separate field. The fact that Course Length is rendered as a radio option is, in my opinion, a concern of the UI only which should be responsible for translating between its chosen presentation and the orthogonal enrichment data. That is:

For "One year" radio option, `CourseLength` should be set to "One year"

For "Two years" radio option, `CourseLength` should be set to "Up to two years"

For Other opition, `CourseLength` should be sent to the content of the rendered text box.

Adding the separation of the default and custom options to the API Model would make UI concerns bleed into the API

dankmitchell:
I agree with Daniel

timabell:
Agree with Daniel. Reject
:-1:

defong:
looks like its decided

MartinBentley:
I originally went down this route for the all reasons that you set out as it seemed clearly the best approach, however after some careful consideration I changed my mind for two reasons:
1) It is a custom input by the user rather than a hard-coded option (all set up in a dictionary as key-value pairs).
2) Adding a new field extends the existing functionality rather than modifying it (SOLID). to modify the existing field I would be diving into the dictionary specific code that generates and converts all the existing options which may break something. By simply adding a new field I can just create new code blocks in the UI partial views to expose the new field only when "Other" is selected ... a more "Semantic" approach. 
Anyway the Ui is nearly complete apart from some styling so I thought the solution would be to finish this up "as is" and then card up the suggested approach if it is still considered the best way.

MartinBentley:
We should not merge this PR. The ui is not using this extra field

PR #128, opened 2018-09-06
Cascade email updates to org-user [wip]

timabell:
### Context

Was doing something tenuously related and accidentally created this small improvement.

Without this you can't update any email addresses because of the FK.
This is a half-way house between that and switching to Id based FKs (which is a bigger change).

With this you can now issue an `update mc_user set email =...` and it works instead of throwing a FK error.

Switch on cascade update/delete on the orguser-user FK.


timabell:
Not quite ready yet, need to work around EF

PR #129, opened 2018-09-07
Populate email in identity from mc_user instead of sign-in claims [wip]

timabell:
This makes our system's knowledge of the email internally consistent.

This fixes an invalid operation exception caused by the terms filter looking for a user by email address when their sign-in email address didn't match the one in mc_users. (We could still find the user because they had previously signed in with their old address and we had their subjectId logged).

### DIFF
--- a/src/ManageCourses.Api/Services/Users/UserService.cs
+++ b/src/ManageCourses.Api/Services/Users/UserService.cs
@@ -26,7 +26,7 @@ namespace GovUk.Education.ManageCourses.Api.Services.Users
         }
 
         /// <inheritdoc />
-        public async Task UserSignedInAsync(string accessToken, JsonUserDetails userDetails)
+        public async Task<McUser> UserSignedInAsync(string accessToken, JsonUserDetails userDetails)
         {
             var mcUser = await _context.McUsers.SingleOrDefaultAsync(u => u.SignInUserId == userDetails.Subject);
             if (mcUser == null)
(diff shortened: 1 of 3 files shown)


src/ManageCourses.Api/Services/Users/UserService.cs
```diff
@@ -59,6 +59,7 @@ public async Task UserSignedInAsync(string accessToken, JsonUserDetails userDeta
 
             _context.Save();
             SendWelcomeEmail(mcUser);
+            return mcUser;
```
defong: I think this was the original dawn of time approach that you disliked right?

timabell:
[Manage Courses API 500 error on prod - changed email in sign-in](https://trello.com/c/1DRswJMM/252-manage-courses-api-500-error-on-prod-changed-email-in-sign-in)

src/ManageCourses.Api/Services/Users/UserService.cs
timabell: Probably :-)

timabell: We can take it back out again when we switch to Id as a FK

timabell:
Waiting for https://github.com/DFE-Digital/manage-courses-api/pull/131 to go in first.

timabell:
I have no idea how to write a sensible test for the failure mode we had. At least not on a friday afternoon anyway...

the-daniel-rothig:
Tim's work on https://github.com/DFE-Digital/manage-courses-api/pull/131 supersedes this PR. Closing

PR #130, opened 2018-09-07
Don't include email in exception

timabell:

### DIFF
--- a/src/ManageCourses.Api/ActionFilters/AcceptTermsFilter.cs
+++ b/src/ManageCourses.Api/ActionFilters/AcceptTermsFilter.cs
@@ -48,7 +48,7 @@ namespace GovUk.Education.ManageCourses.Api.ActionFilters
             }
             if (users.Count == 0)
             {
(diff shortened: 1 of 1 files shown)


timabell:
[capturing email in logs (bad)](https://trello.com/c/AZpzcV9S/268-capturing-email-in-logs-bad)

benilovj:
Is the UUID from DfE Sign-in available in this context? Because you could log that instead and we still have a way to figure out who it is if needed...

defong:
based on 

https://github.com/DFE-Digital/manage-courses-api/blob/9f1b676443c461b51c08f3c8d309ab153936b900/src/ManageCourses.Api/Middleware/BearerTokenHandler.cs#L63

it be something like

`var subjectId = User.Claims.Single(x => x.ValueType == ClaimTypes.NameIdentifier);`

timabell:
I think we should punt that into the backlog for a broader fix https://trello.com/c/9PK6MBvd/276-add-sign-in-subject-id-to-app-insights-info

PR #137, opened 2018-09-11
Fix api status codes for unauth'd users.

timabell:
Fail was causing 404 (redirect to non-existent login page we think).
Use NoResult instead to give 401.

### Context
It was all broken. UI was giving 404 instead of the "we don't know who you are" page.

Fix it. Break the tests. >_<


timabell:
[404 and 5XX errors shouldn't have their own URLs in Manage courses](https://trello.com/c/QaRjynru/232-404-and-5xx-errors-shouldnt-have-their-own-urls-in-manage-courses)

timabell:
Tests failing in setup because https://trello.com/c/MjNZSdMt/55-bearer-token-handler-called-for-api-key-endpoints-timebox-30-mins

PR #140, opened 2018-09-12
Don't hit sign-in endpoints for api-key protected endpoints

timabell:
### Context

API was hitting sign-in end point for every api-key based request causing extra load, errors and confusion.

* Remove default schemes
* Specify scheme on every endpoint

### DIFF
--- a/src/ManageCourses.Api/Controllers/AcceptTermsController.cs
+++ b/src/ManageCourses.Api/Controllers/AcceptTermsController.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Linq;
 using GovUk.Education.ManageCourses.Api.ActionFilters;
+using GovUk.Education.ManageCourses.Api.Middleware;
 using GovUk.Education.ManageCourses.Domain.DatabaseAccess;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
@@ -17,7 +18,7 @@ namespace GovUk.Education.ManageCourses.Api.Controllers
             this.context = context;
         }
 
-        [Authorize]
+        [Authorize(AuthenticationSchemes = BearerTokenDefaults.AuthenticationScheme)]
         [ExemptFromAcceptTerms]
         [HttpPost]        
         [Route("accept")]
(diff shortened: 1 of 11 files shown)


src/ManageCourses.Api/Controllers/AcceptTermsController.cs
```diff
@@ -17,7 +18,7 @@ public AcceptTermsController(IManageCoursesDbContext context)
             this.context = context;
         }
 
-        [Authorize]
+        [Authorize(AuthenticationSchemes = BearerTokenDefaults.AuthenticationScheme)]
```
defong: Can we create a class that inherits from `Authorize` ie `BearerTokenAuthorize` and its counterpart ``BearerTokenApiKeyAuthorize``

So it shows up as a hard choice made therefore usage of `Authorize` can be quickly discovered quickly

src/ManageCourses.Api/Startup.cs
```diff
@@ -52,17 +52,15 @@ public void ConfigureServices(IServiceCollection services)
...
-                options.DefaultAuthenticateScheme = BearerTokenDefaults.AuthenticationScheme;
-                options.DefaultChallengeScheme = BearerTokenDefaults.AuthenticationScheme;
-            }).AddBearerToken(options =>
-            {
-                options.UserinfoEndpoint = Configuration["auth:oidc:userinfo_endpoint"];
-            }).AddBearerTokenApiKey(options =>
-            {
-                options.ApiKey = Configuration["api:key"];
-            });
+            services.AddAuthentication()
```
defong:
```
services.AddAuthentication(
 // And explain it here of the available choices
 // As AuthorizeAttribute was designed to use a default that we no longer set.
)
```

src/ManageCourses.Api/Controllers/AcceptTermsController.cs
timabell: done. good call :+1:

src/ManageCourses.Api/Startup.cs
timabell: Not sure what to write here, suggestions?

defong: use one of the following?

defong: BearerTokenAuthorize and its counterpart BearerTokenApiKeyAuthorize as `Authorize` is dead?

timabell: done

PR #142, opened 2018-09-13
Move course mapper into api

MartinBentley:
### Context
We need to publish a course into Search and Compare.
Part of this is to move the Course mapper into the Api client. This is the Pr to do this
The Course mapper has been put into the Api client project


timabell:
Approved as-is because it's a lift-n-shift from elsewhere

PR #144, opened 2018-09-13
Config validation to pin down cause of flapping test

timabell:
### Context

One of our tests has been intermittently failing with a null ref.
I have a hunch it's something config related so this PR tightens up the config handling.

* Strongly typed config
* Validate config earlier


src/ManageCourses.Api/McConfig.cs
```diff
@@ -11,23 +12,50 @@ public McConfig(IConfiguration configuration)
             _configuration = configuration;
         }
 
+        public void Validate()
```
defong: It looks wrong, feels wrong, so surely its wrong

PR #148, opened 2018-09-13
Another take on config validation

timabell:
Also don't validate password as it's not required

### Context
PR feedback that missed the boat
https://github.com/DFE-Digital/manage-courses-api/pull/144#discussion_r217405470


timabell:
Rebased before merging

PR #156, opened 2018-09-21
Fix null reference exception in AdminController

the-daniel-rothig:
During a manual access request, the retireval of the reference user was
incomplete. We didn't include their associated organisation so,
when it came to adding the new user to these organsiations, the org
collection could not be accessed.


tvararu:
I've updated this to handle an edge-case where targetEmail or requesterEmail is ommited, which would result in a 500.

@timabell can you please re-review?

PR #160, opened 2018-09-25
Fix how we select subjects for courses.

the-daniel-rothig:
We were using all subjects of all courses of a course's institution
when retrieving course subjects. That's obviously not what should happen.

### DIFF
--- a/src/ManageCourses.Api/Mapping/CourseLoader.cs
+++ b/src/ManageCourses.Api/Mapping/CourseLoader.cs
@@ -53,7 +53,7 @@ namespace GovUk.Education.ManageCourses.Api.Mapping
                 returnCourse.ProgramType = organisationCourseRecord.ProgramType;
                 returnCourse.ProfpostFlag = organisationCourseRecord.ProfpostFlag;
                 returnCourse.StudyMode = organisationCourseRecord.Studymode;
-                var subjects = organisationCourseRecord.UcasInstitution.UcasCourseSubjects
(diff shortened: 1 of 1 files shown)


src/ManageCourses.Api/Mapping/CourseLoader.cs
```diff
@@ -53,7 +53,7 @@ public Course LoadCourse(IEnumerable<UcasCourse> courseRecords, IEnumerable<Ucas
                 returnCourse.ProgramType = organisationCourseRecord.ProgramType;
                 returnCourse.ProfpostFlag = organisationCourseRecord.ProfpostFlag;
                 returnCourse.StudyMode = organisationCourseRecord.Studymode;
-                var subjects = organisationCourseRecord.UcasInstitution.UcasCourseSubjects
+                var subjects = organisationCourseRecord.CourseCode.UcasCourseSubjects
```
timabell: > ...CourseCode.UcasCourseSubjects

Is `CourseCode` a strangely named property because that doesn't make a lot of sense!

timabell: Yes, yes it is, and I created it.
It's the join table I added to the ucas data to give us unique inst/course code combinations. I couldn't call it UcasCourse because that's already there and is actually a ucas-course-campus thing.

timabell: Made a card should we want to clean up https://trello.com/c/NrOqnHQ1/377-rename-ucas-tables

PR #164, opened 2018-09-25
Merge in ucas org contact details even if there's no enrichment

timabell:
### Context

Loading an un-enriched org showed no contact details at all, not even the ones available from the ucas data. They would then magically appear when you saved the other org enrichment fields.

* Return merged enrichment even if there was no existing enrichment record

### Guidance to review

Note that this means that the ucas contact details will be copied into the enrichment data the first time the user hits save on any of the org enrichment data. Jake said this is fine.


timabell:
[ContactDetails: get contact details either from ucas or from enrichment](https://trello.com/c/Sr2qhHHh/285-contactdetails-get-contact-details-either-from-ucas-or-from-enrichment)

PR #175, opened 2018-09-28
Extra flag for send to search and compare

MartinBentley:
### Context
A new feature flag has been introduced into the manage courses ui to switch off sending to search and compare when a course is published.
The Publish controller end point was previously called "PublishToSearchAndCompare" however it was both changing the course status to "Published" and also sending the course across to search and compare. The endpoint has now been renamed to PublishCourse as it now has an extra boolean parameter that determines whether or not to send it across to S/C


src/ManageCourses.Api/Controllers/PublishController.cs
```diff
@@ -31,13 +31,16 @@ public PublishController(IDataService dataService, IEnrichmentService enrichment
         [ProducesResponseType(typeof(bool), 200)]
         [ProducesResponseType(400)]
         [ProducesResponseType(404)]
-        public async Task<ActionResult> PublishToSearchAndCompare(string instCode, string courseCode)
```
timabell: this should be configured from an environment variable rather than a parameter

Soundman32: Should this function be called PublishCourseAsync?

MartinBentley: Why would we want the setting in two places?  The Ui has the new feature flag. This setting is passed then into the Api via the parameter.  This endpoint doing is doing 2 things here: 1 changing the Status and 2. Sending to SearchAndCompare. Previously we had no control over this. Also the return value from the SearchAndCompare service is totally ignored so it's really a 'fire and forget' action. The parameter seems to me the best option! Happy to be overruled though.

MartinBentley: Agreed. I'll change it

PR #181, opened 2018-10-02
Make correct course qualification available

the-daniel-rothig:
### Context

Correct qualifications in Manage courses UI

Previously the qualification was determined as part of CourseMapper,
which only has an effect during preview and publish.

This pushes the qualification logic down to CourseLoader, making the
qualification available to everything that uses `Course`. This will
allow manage-courses-ui to show "Type" strings consistent with
Search and Compare.


src/ManageCourses.Api/Model/Course.cs
```diff
@@ -21,7 +21,9 @@ public class Course
         public string AccreditingProviderName { get; set; }
         public string Subjects { get; set; }
         public string StudyMode { get; set; }
-        public string AgeRange { get; set; }
+        public string AgeRange { get; set; }        
+        public CourseQualification Qualification { get; set; }
+        public string TypeDescription { get; set; }
```
timabell: xmldoc with the kind of thing you'd expect to see in here

eg "qual(s) + full/part time + salary + apprenticeship"

src/ManageCourses.Api/Model/CourseQualification.cs
```diff
@@ -0,0 +1,11 @@
+namespace GovUk.Education.ManageCourses.Api.Model
+{
+    public enum CourseQualification
+    {
+        Qts,
```
timabell: xml doc with full names
"Qualified Teacher Status"
while we still remember what they are...

src/ManageCourses.Api/Services/Data/DataService.cs
```diff
@@ -19,13 +19,15 @@ public class DataService : IDataService
...
+        private readonly IPgdeWhitelist _pgdeWhitelist;
 
-        public DataService(IManageCoursesDbContext context, IEnrichmentService enrichmentService, IDataHelper dataHelper, ILogger<DataService> logger)
+        public DataService(IManageCoursesDbContext context, IEnrichmentService enrichmentService, IDataHelper dataHelper, ILogger<DataService> logger, IPgdeWhitelist pgdeWhitelist)
         {
             _context = context;
             _enrichmentService = enrichmentService;
             _dataHelper = dataHelper;
             _logger = logger;
+            this._pgdeWhitelist = pgdeWhitelist;
```
timabell: `this` isn't needed

PR #185, opened 2018-10-09
CourseLoader to prefer running locations

the-daniel-rothig:
### Context

Source: Zendesk

Users of UCAS weblink can only edit the data of running locations. For
this reason, when e.g. the Accrediting Training Provider changes,
there will be an inconsistency between previously discontinued locations
and current locations.

So take data from running courses if possible.

### Guidance to review

See tests

### DIFF
--- a/tests/ManageCourses.Tests/UnitTesting/Mapping/CourseLoaderTests.cs
+++ b/tests/ManageCourses.Tests/UnitTesting/Mapping/CourseLoaderTests.cs
@@ -39,6 +39,41 @@ namespace GovUk.Education.ManageCourses.Tests.UnitTesting.Mapping
             res.StartDate.Should().BeNull();
         }
 
+        [Test]
+        public void RunningLocationsArePreferred()
+        {
+            var sut = new CourseLoader();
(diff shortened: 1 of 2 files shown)


tests/ManageCourses.Tests/UnitTesting/Mapping/CourseLoaderTests.cs
```diff
@@ -39,6 +39,41 @@ public void TolerateBlankStartDate()
...
+            res.AccreditingProviderId.Should().Be("RIGHT_ACC");
+            res.InstCode.Should().Be("RIGHT_INST");
+        }
+
+        [Test]
+        public void FullySuspendedCoursesWorkStill()
+        {            
+            var sut = new CourseLoader();
+            var loc1 = GetBlankUcasCourse();
+            loc1.Status = "S";
```
defong: what is `S`?

the-daniel-rothig: "suspended". There are also *D* for discontinued, *N* for new, and of course *R* for running

PR #186, opened 2018-10-09
This is why duplication is a bad thing

the-daniel-rothig:
### Context

In the UI we had accounted for apprenticeships in our type description.
In the API we hadn't yet. When removing duplication we lost coverage of
this case.

This commit reintroduces it.


defong:
Can we have a test for it?

PR #183, opened 2018-10-03
Wired up publishing to new SearchAndCompare endpoint.

MartinBentley:
### Context
The SearchAndCompare publish course endpoint now takes a list of course to publish
### Changes proposed in this pull request
The PublishService now has two methods for publish course and publish courses

PR #188, opened 2018-10-10
Built wrappers

MartinBentley:
### Context
This is a change that was requested during the PR review of the publishing story:
https://github.com/DFE-Digital/manage-courses-api/pull/183.
The deal was that we would let it go and create a separate pr. And this is it!
To create two wrapper methods for each existing method in the enrichment service that gets an organisation or course enrichment.

### DIFF
--- a/src/ManageCourses.Api/Services/EnrichmentService.cs
+++ b/src/ManageCourses.Api/Services/EnrichmentService.cs
@@ -20,68 +20,23 @@ namespace GovUk.Education.ManageCourses.Api.Services
             _converter = new EnrichmentConverter();
         }
 
-        /// <summary>
-        /// Gets the latest enrichment record if any regardless of the status.
-        /// If there is no contact data then it will be pre-populated from the Ucas data.
-        /// </summary>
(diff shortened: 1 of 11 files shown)


src/ManageCourses.Api/Services/EnrichmentService.cs
```diff
@@ -20,68 +20,23 @@ public EnrichmentService(IManageCoursesDbContext context)
...
-                    enrichmentModel.Address3 = ucasInst.Addr3;
-                    enrichmentModel.Address4 = ucasInst.Addr4;
-                    enrichmentModel.Postcode = ucasInst.Postcode;
-                }
-            }
-
-            return enrichmentGetModel;
+            return GetInstitutionEnrichment(instCode, email, false);
         }
+        public UcasInstitutionEnrichmentGetModel GetInstitutionEnrichmentForPublish(string instCode, string email)
```
defong: Could of done that

public UcasInstitutionEnrichmentGetModel GetInstitutionEnrichmentForPublish(string instCode, string email) =>
            return GetInstitutionEnrichment(instCode, email, true);

PR #189, opened 2018-10-11
Amended travis to use 2.1 same as UI rather then 2.1.300

defong:
### Context
travis dotnet version
Amended travis to use 2.1 same as UI rather then 2.1.300


defong:
[Wondering why staging `manage-courses-api` is broken](https://trello.com/c/SrHVmYiE/483-wondering-why-staging-manage-courses-api-is-broken)

PR #192, opened 2018-10-16
Ucas import efficiency

the-daniel-rothig:
### Context

We have been investigating improvements to the import process from ManageCourses to SearchAndCompare. One thing that we discovered is how to use DB transactions more effectively to make transitions more atomic.

There are two commits on offer here ;

`d6498884` replaces a handcoded, brittle rollback algorithm with the database one. 
`5c4651d` goes one step further and merges all updates, which previously were split on a per-institution basis

### Guidance to review

This is a nonfunctional change so no new tests. Local tests confirm performance improvements.

* Without this PR, locally the import takes about 6 minutes (n.b. on production this takes about 90 minutes)
* With the first commit, locally the import takes about 5.5 minutes with improved referential integrity
* With the second commit, locally the import takes about 1.5 minutes, but includes a behavioural change: any malformed import request will now cancel the entire import, see failing test. I suspect that the gains on production will be higher due to lowered connection overheads to the database

I'd like some input on whether we should include the second commit (in which case I will fix up the test) or not (in which case I remove it)

My inclination is to include both as the per-institution split is a bit anachronistic but want to hear other opinions

### DIFF
--- a/src/ManageCourses.Api/Services/Data/DataService.cs
+++ b/src/ManageCourses.Api/Services/Data/DataService.cs
@@ -45,62 +45,34 @@ namespace GovUk.Education.ManageCourses.Api.Services.Data
             int processed = 0;
             foreach (var inst in allInstitutions)
             {
-                // note: this does not perform a delete of existing institutions
-
-                var instRes = AttemptTransaction(inst.InstCode, "upsert institution entity", () => UpsertInstitution(inst));
-
-                var deleted = AttemptTransaction(inst.InstCode, "delete for replace", () => DeleteForInstitution(inst.InstCode));
-                var added = AttemptTransaction(inst.InstCode, "insert", () => AddForInstitution(inst.InstCode, payload));
-
-                if (added == null && deleted != null)
+                using (var transaction = (_context as DbContext).Database.BeginTransaction())
                 {
-                    //deletion succeeded but adding didn't... attempt to recover by re-adding deleted content
-                    AttemptTransaction(inst.InstCode, "restore", () => AddForInstitution(inst.InstCode, deleted));
+                    try 
+                    {
+                        UpsertInstitution(inst);
+                        _context.Save();
+                        DeleteForInstitution(inst.InstCode);
+                        _context.Save();
+                        AddForInstitution(inst.InstCode, payload);
+                        _context.Save();
+                        transaction.Commit();
+                    }
+                    catch (Exception e)
+                    {
+                        transaction.Rollback();
+                        _logger.LogError(e, $"UCAS import failed to update institution {inst.InstFull} [{inst.InstCode}]");
+                    }
                 }
                 if (++processed % 100 == 0)
                 {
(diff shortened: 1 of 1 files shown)


defong:
maybe add a third commit as its always a charm

replace 
`_context.Addxxx(singleItem)`

with 
`_context.xxx.AddRange(listOfItems)`

src/ManageCourses.Api/Services/Data/DataService.cs
```diff
@@ -42,109 +42,64 @@ public void ProcessUcasPayload(UcasPayload payload)
...
-
-                var deleted = AttemptTransaction(inst.InstCode, "delete for replace", () => DeleteForInstitution(inst.InstCode));
-                var added = AttemptTransaction(inst.InstCode, "insert", () => AddForInstitution(inst.InstCode, payload));
-
-                if (added == null && deleted != null)
+                try 
                 {
-                    //deletion succeeded but adding didn't... attempt to recover by re-adding deleted content
-                    AttemptTransaction(inst.InstCode, "restore", () => AddForInstitution(inst.InstCode, deleted));
+                    foreach(var inst in allInstitutions)
```
defong: Change `UpsertInstitution` to `UpsertInstitutions`
Return a list of Institution that was either newly added or updated as `iAmNowInContext`
Looks like its missing 
`_context.Save()`

Anyways then in `AddUcasData(payload, iAmNowInContext)`

And do the match up of institution via `ref `

src/ManageCourses.Api/Services/Data/DataService.cs
```diff
@@ -42,109 +42,64 @@ public void ProcessUcasPayload(UcasPayload payload)
...
+                    _context.Save();
+
+                    
+                    transaction.Commit();
                 }
-                if (++processed % 100 == 0)
+                catch (Exception)
                 {
-                    _logger.LogInformation($"Upserted {processed} institutions so far");
+                    transaction.Rollback();
```
defong: log i have failed please

timabell:
Seems reasonable, +1 to Sau's comments.

The only thing that occurs to me is that doing everything in one transaction means that if ucas feed us data that causes a failure in one institution then with this change the whole thing would stop working and no updates would happen till we've fixed it up. Having said that I'm not aware that's ever happened so I'd take the speed advantage for sure.

defong:
Coming to think of it if we have a list of institution
we can further optimise it by making 
`AddUcasData(payload, iAmNowInContext)`

slice and dice in parallel

timabell:
[Periodic automatic publishing of all courses from Publish to Search](https://trello.com/c/YDas1pT0/289-periodic-automatic-publishing-of-all-courses-from-publish-to-search)

the-daniel-rothig:
Decided to revert to one-transaction-per-institution. Reason: I don't like the idea to lock a big part of the database into one huge transaction, it's a recipe for locking and deadlocking nightmares.

PR #194, opened 2018-10-17
Exporter util

timabell:
### Context
This wasn't intended to actually be put into master but times have changed so now here it comes!

We are going to turn this into a webjob and run it alongside the manage api. Before I do the work to turn it into a web job I'd like the current state of this (for better or worse) to be merged into `master`.

We can schedule this now because both sides are now transactional thanks to https://github.com/DFE-Digital/search-and-compare-api/pull/82 and #192

Adding of a quick-n-dirty command-line util that takes data straight from the manage db, generates a 100Mb json file, and then a bit of nodejs to post it to search and compare.

### Guidance to review
Don't look too closely.


src/ManageCourses.Api/Mapping/SubjectMapper.cs
```diff
@@ -172,11 +172,11 @@ static SubjectMapper()
             ucasUnexpected = new string[]
             {
                 "construction and the built environment",
-                "history of art",
+                //"history of art",
```
timabell: merge error?

src/ManageCourses.Api/Mapping/SubjectMapper.cs
```diff
@@ -172,11 +172,11 @@ static SubjectMapper()
...
                 "home economics",
                 "hospitality and catering",
                 "personal and social education",
-                "philosophy",
+                //"philosophy",
```
timabell: merge error?

timabell:
@MartinBentley yes, we do :-)

PR #201, opened 2018-10-24
Big database refactor

the-daniel-rothig:
### Context

We want for our database schema to be cleaner, have better referential integrity, and be decoupled from pre-existing concepts.

This is a big restructure.
* The `ucas_*` tables are gone, replaced with our version of `course` and `institution`
* The mapping from the UCAS schema to ours now happens before import
* The `course_codes` table was removed as its records now have equivalent meaning with `course`
* the UCAS importer is now deployed alongside manage-courses-api, rendering the manage-courses-ucas-importer repo redundant
* the UCAS importer talks straight to the `DbContext` without going via the API (eliminating timeouts and a fair amount of boilerplate code)
* We use DB-generated `integers` for keys now [as Codd intended](https://i.pinimg.com/originals/db/d9/3d/dbd93d7a2206bd44bb83f1d8b515e50c.png)
* Remove the (now-redundant) `Mc` prefixes in affected tables, columns, types, properties and files
* A few naming inconsistencies such as `Addr` vs `Address`, `InstFull` vs `InstName` and `Provider` vs `Institution` have been eliminated as well

### Guidance to review

I managed to keep all tests alive, but since this is a substantial and all-encompassing change, a lot of manual testing should be done as well.

The data imported by UcasCourseImporter gets dropped as part of the database migration as mapping it is impractical. So the UCAS importer needs to be run again (in production this will incur some downtime).

Reference data on the other hand will be lovingly mapped across.


the-daniel-rothig:
Sounds good to me!

PR #204, opened 2018-10-25
hotfix: subjects now mean our subjects, not ucas_subjects

the-daniel-rothig:


fofr:
What's the effect of this PR?

fofr:
My concern is that we do still need to show UCAS subjects here:

screen shot 2018-10-25 at 13 48 27

PR #206, opened 2018-10-26
Fix two indexing issues

the-daniel-rothig:
### Context

Follow up from the big ol' database refactor

- Reinstate SubjectCode indexing
- Change the nonclustered index on Course to use CourseCode, which aids
  when requesting courses by ucas codes


timabell:
Does more than the title implies

PR #207, opened 2018-10-26
Reintroduce SubjectMapper into CourseMapper

the-daniel-rothig:
### Context

Leftover from the big ol' database refactor

We store ucas subjects in the database which need to be converted to
DfE subjects on export. This behaviour was removed temprarily when
we changed what we stored in the DB to DfE subjects.

This undoes that.

### DIFF
--- a/src/ManageCourses.Api/Mapping/CourseMapper.cs
+++ b/src/ManageCourses.Api/Mapping/CourseMapper.cs
@@ -12,6 +12,8 @@ namespace GovUk.Education.ManageCourses.Api.Mapping
 {
     public class CourseMapper : ICourseMapper
     {
+        private readonly SubjectMapper subjectMapper = new SubjectMapper();
+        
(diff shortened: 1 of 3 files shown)


src/ManageCourses.Api/Mapping/CourseMapper.cs
```diff
@@ -12,6 +12,8 @@ namespace GovUk.Education.ManageCourses.Api.Mapping
 {
     public class CourseMapper : ICourseMapper
     {
+        private readonly SubjectMapper subjectMapper = new SubjectMapper();
```
defong: Inject this please

the-daniel-rothig: Why? No dependencies, no polymorphism, so DI could never change behaviour nor simplify the code

the-daniel-rothig:
You remind me of my geography teacher ;)

PR #214, opened 2018-11-01
Flush importer logs

timabell:
### Context

We were not seeing the completion message in the azure portal logs.

Do as suggested in https://github.com/serilog/serilog-sinks-applicationinsights#how-when-and-why-to-flush-messages-manually


timabell:
[Fixup prod](https://trello.com/c/tMO30eOw/549-fixup-prod)

timabell:
[Flush application insights data before app shutdown](https://trello.com/c/eAsAEq3B/170-flush-application-insights-data-before-app-shutdown)

PR #221, opened 2018-11-14
Add offline page

MartinBentley:


MartinBentley:
It does seem strange to deliver an htm page from an api but it gives us a mechanism to easily shut down the service.

PR #222, opened 2018-11-14
Ping

timabell:
### Context
Can't currently tell if api is up without front end auth.

Add an endpoint for a quick check of api + db health.


timabell:
[ping for manage-api](https://trello.com/c/O0GCelKh/44-ping-for-manage-api)

PR #225, opened 2018-11-16
Capture logs at startup

timabell:
As per https://github.com/DFE-Digital/search-and-compare-api/pull/89


timabell:

PR #227, opened 2018-11-21
Enable some retrying for postgres

timabell:
Turn on the built-in retry for postgres. Will retry timeout exceptions that we saw in prod under poor network conditions. Will not retry all error conditions, for that see https://github.com/DFE-Digital/manage-courses-api/pull/228

### DIFF
--- a/src/ManageCourses.Api/Startup.cs
+++ b/src/ManageCourses.Api/Startup.cs
@@ -1,4 +1,6 @@
-﻿using System.Net.Http;
+﻿using System;
+using System.Collections.Generic;
+using System.Net.Http;
 using System.Net.Http.Headers;
 using System.Reflection;
 using GovUk.Education.ManageCourses.Api.ActionFilters;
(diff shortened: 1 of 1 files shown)


timabell:

src/ManageCourses.Api/Startup.cs
```diff
@@ -51,9 +53,15 @@ public void ConfigureServices(IServiceCollection services)
                 .AddDbContext<ManageCoursesDbContext>(
                     options =>
                     {
+                        const int maxRetryCount = 3;
+                        const int maxRetryDelaySeconds = 5;
+
+                        var postgresErrorCodesToConsiderTransient = new List<string>(); // ref: https://github.com/npgsql/Npgsql.EntityFrameworkCore.PostgreSQL/blob/16c8d07368cb92e10010b646098b562ecd5815d6/src/EFCore.PG/NpgsqlRetryingExecutionStrategy.cs#L99
```
the-daniel-rothig: is this needed?

timabell: yes it doesn't have an overload without that param <sadface>

PR #229, opened 2018-11-23
Include migration exception in message string

timabell:


src/ManageCourses.Api/Startup.cs
```diff
@@ -191,7 +191,8 @@ private void Migrate(ManageCoursesDbContext dbContext)
                     {
                         delayMs = maxDelayMs;
                     }
-                    _logger.LogError($"Failed to apply EF migrations. Attempt {migrationAttempt} of ∞. Waiting for {delayMs}ms before trying again.", ex);
+                    // exception included in message string because app insights isn't showing the messages and kudo log stream only shows the message string.
```
defong: I think you need something like this

defong: https://github.com/DFE-Digital/manage-courses-ui/blob/a1b03b3073e78f4d51fab86ca82ee798758093ab/src/ui/ActionFilters/McExceptionFilter.cs#L47

defong: You swallowed the exception, nothing bubbles up...

Makes sense that its not on show

timabell: nice one

timabell:

src/ManageCourses.Api/Startup.cs
timabell: done

PR #233, opened 2018-11-30
Fix access requests with retry

timabell:
### Context

Access requests have been broken in prod since we introduced the retry strategy in #227 

https://docs.microsoft.com/en-gb/azure/architecture/best-practices/retry-service-specific#sql-database-using-entity-framework-core

### Guidance to review

Turn off whitespace diff

### DIFF
--- a/src/ManageCourses.Api/Services/AccessRequests/AccessRequestService.cs
+++ b/src/ManageCourses.Api/Services/AccessRequests/AccessRequestService.cs
@@ -24,6 +24,7 @@ namespace GovUk.Education.ManageCourses.Api.Services.AccessRequests
 
         public void LogAccessRequest(AccessRequest request, string requesterEmail)
         {
+            // Strategy is required for compatibility with EnableRetryOnFailure in Startup. Ref https://docs.microsoft.com/en-gb/azure/architecture/best-practices/retry-service-specific#sql-database-using-entity-framework-core
             var strategy = RawContext.Database.CreateExecutionStrategy();
             strategy.Execute(() =>
             {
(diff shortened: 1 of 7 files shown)


src/ManageCourses.Api/Services/AccessRequests/AccessRequestService.cs
```diff
@@ -20,48 +20,55 @@ public AccessRequestService(IManageCoursesDbContext context, IAccessRequestEmail
...
         }
 
+        public DbContext RawContext => (DbContext)_context;
+
         public void LogAccessRequest(AccessRequest request, string requesterEmail)
         {
-            using (var transaction = ((DbContext)_context).Database.BeginTransaction())
+            // Strategy is required for compatibility with EnableRetryOnFailure in Startup. Ref https://docs.microsoft.com/en-gb/azure/architecture/best-practices/retry-service-specific#sql-database-using-entity-framework-core
+            var strategy = RawContext.Database.CreateExecutionStrategy();
+            strategy.Execute(() =>
```
defong: Can we not expose it in dbcontext interface?
Rather than this casting? `DbContext RawContext => (DbContext)_context`

timabell:
Tests are failing because the change to the tests affected all the tests, but some of them test the ucas importer which doesn't use the retry system.

src/ManageCourses.Api/Services/AccessRequests/AccessRequestService.cs
timabell: Yeah I guess so

PR #250, opened 2019-01-17
Rename timestamps

timabell:
### Context
Rails has expectations for timestamp field names. Refactor to match that.

* Rename
* Migrations
* Code formatting

### Guidance to review

This produced no diff in the swagger json so should have no effect on the api surface, resulting in no need to change the UI.

### DIFF
--- a/src/ManageCourses.Api/Mapping/EnrichmentConverter.cs
+++ b/src/ManageCourses.Api/Mapping/EnrichmentConverter.cs
@@ -68,7 +68,7 @@ namespace GovUk.Education.ManageCourses.Api.Mapping
             enrichmentToReturn.UpdatedTimestampUtc = source.UpdatedAt;
             enrichmentToReturn.CreatedByUserId = source.CreatedByUser.Id;
             enrichmentToReturn.UpdatedByUserId = source.UpdatedByUser.Id;
-            enrichmentToReturn.LastPublishedTimestampUtc = source.LastPublishedTimestampUtc;
+            enrichmentToReturn.LastPublishedTimestampUtc = source.LastPublishedAt;
             enrichmentToReturn.Status = source.Status;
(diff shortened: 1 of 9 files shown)


timabell:
[Refactor created and updated timestamps on provider enrichment](https://trello.com/c/1Ld4Hjgv/868-refactor-created-and-updated-timestamps-on-provider-enrichment)

src/ManageCourses.Api/Mapping/EnrichmentConverter.cs
```diff
@@ -64,11 +64,11 @@ public UcasProviderEnrichmentGetModel Convert(ProviderEnrichment source)
...
 
             enrichmentToReturn.EnrichmentModel = enrichmentModel;
-            enrichmentToReturn.CreatedTimestampUtc = source.CreatedTimestampUtc;
-            enrichmentToReturn.UpdatedTimestampUtc = source.UpdatedTimestampUtc;
+            enrichmentToReturn.CreatedTimestampUtc = source.CreatedAt;
+            enrichmentToReturn.UpdatedTimestampUtc = source.UpdatedAt;
             enrichmentToReturn.CreatedByUserId = source.CreatedByUser.Id;
             enrichmentToReturn.UpdatedByUserId = source.UpdatedByUser.Id;
-            enrichmentToReturn.LastPublishedTimestampUtc = source.LastPublishedTimestampUtc;
+            enrichmentToReturn.LastPublishedTimestampUtc = source.LastPublishedAt;
```
misaka: What about just `published_at`?

misaka: Just had a look with @timabell and this attribute is a bit quirky, so we'll leave it as-is and consider refactoring at a later time.

PR #251, opened 2019-01-17
Provider timestamps

timabell:
### Context
Providing incremental data feed for UCAS apply system.

Populate the new timestamp fields during import

### DIFF
--- a/src/ManageCourses.Domain/Migrations/20190117145424_ProviderTimestamps2.cs
+++ b/src/ManageCourses.Domain/Migrations/20190117145424_ProviderTimestamps2.cs
@@ -1,0 +1,54 @@
+﻿using System;
+using Microsoft.EntityFrameworkCore.Migrations;
+
+namespace GovUk.Education.ManageCourses.Domain.Migrations
+{
+    public partial class ProviderTimestamps2 : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.AddColumn<DateTime>(
+                name: "created_at",
+                table: "site",
+                nullable: false,
+                defaultValueSql: "now() at time zone 'utc'"); // https://stackoverflow.com/questions/16609724/using-current-time-in-utc-as-default-value-in-postgresql/16610360#16610360
+
+            migrationBuilder.AddColumn<DateTime>(
+                name: "updated_at",
+                table: "site",
+                nullable: false,
+                defaultValueSql: "now() at time zone 'utc'");
+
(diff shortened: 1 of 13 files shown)


timabell:
[Set `created_at` and `updated_at` in importer.](https://trello.com/c/oqRrhsnS/867-set-createdat-and-updatedat-in-importer)

src/ManageCourses.Domain/Migrations/20190117145424_ProviderTimestamps2.cs
```diff
@@ -0,0 +1,54 @@
...
+                name: "created_at",
+                table: "provider",
+                nullable: false,
+                defaultValueSql: "CURRENT_DATE");
+
+            migrationBuilder.AddColumn<DateTime>(
+                name: "updated_at",
+                table: "provider",
+                nullable: false,
+                defaultValueSql: "CURRENT_DATE");
```
timabell: gives just a date. oops

src/ManageCourses.UcasCourseImporter/importer/UcasDataMigrator.cs
```diff
@@ -210,12 +212,15 @@ private Provider UpsertProvider(Provider newValues)
...
                 _context.Providers.Add(newValues);
+                entity.CreatedAt = DateTime.UtcNow;
+                entity.UpdatedAt = DateTime.UtcNow;
                 return newValues;
             }
             else
             {
                 // update
                 entity.UpdateWith(newValues);
+                entity.UpdatedAt = DateTime.UtcNow;
```
misaka: Can we have at least some simple tests for the importer?

defong: https://guides.rubyonrails.org/active_record_basics.html#schema-conventions

does it collides with activerecord?

maybe rename those fields to 

> created_at - Automatically gets set to the current date and time when the record is first created.
updated_at - Automatically gets set to the current date and time whenever the record is updated.

misaka: @defong That's actually why we're doing this, we want to have these columns on the table so that we can use them in the Rails app. Since the Rails app is read-only it won't be updating these, but if it did it would still be the right thing to do, presumably.

timabell:
@defong says: https://docs.microsoft.com/en-us/dotnet/api/system.datetimeoffset?view=netframework-4.7.2

timabell:
everything else is all DateTime so leaving as is

PR #253, opened 2019-01-18
Add more info to deploy output

timabell:
Can't tell which webjob is being deployed in the travis output.

Also make zip less noisy and add some "done" info.


travis-deploy-ucas-importer.sh
```diff
@@ -1,15 +1,16 @@
 #!/usr/bin/env bash
+echo "Deploying ucas-importer..."
 
 deployZip=deploy.zip
-
+echo "Deploy complete."
```
timabell: wtf sau lol

PR #257, opened 2019-01-21
Add accrediting_provider to provider

dankmitchell:
### Context
Is a provider an accrediting_provider?

Add `accrediting_provider` to `provider`

### Guidance to review
It looks like we're already importing the data just not saving it - https://github.com/DFE-Digital/manage-courses-api/blob/master/src/ManageCourses.UcasCourseImporter/xls-reader/XlsReader.cs#L122


src/ManageCourses.Domain/Models/Provider.cs
```diff
@@ -45,6 +46,7 @@ public void UpdateWith(Provider provider)
         public string YearCode { get; set; }
         public string Scitt { get; set; }
         public string SchemeMember { get; set; }
+        public string AccreditingProvider { get; set; }
```
benilovj: May be worth storing this as a boolean, thoughts @misaka ?

dankmitchell: I thought that but as we have changed any other Y/N values to booleans I didn't bother

dankmitchell: i.e. `institution_type` and `copy_form_required`

misaka: I'd definitely prefer these to be boolean, but if it means additional work in the .NET app I would probably hold off just for now. Switching this in Rails should be quite easy.

dankmitchell: Agreed @misaka. For all Y/N data values we should move to boolean but IMO its out of scope for this PR.

PR #258, opened 2019-01-22
Add 'region_code' to provider enrichment

defong:
### Context
`region_code` needs to be seeded to provider enrichment 

Added migration to backfill the region_code in the enrichment for the provider
Added system seeding for first creating the enrichment to ensure that there is a region code
Added system seeding to ensure that when an enrichment is saved that there is a region code

### Guidance to review
Concepts

Existing provider enrichment `region_code` is backfilled from provider `region_code`
There is no need to remove `region_code` from migrations.

New provider enrichment `region_code` uses provider `region_code`
UI has absent knowledge of `region_code` hence system adds it at execution points as required.
That bit needs to be removed when `region_code` logistics are mapped out.


defong:
[Add 'region_code' to provider enrichment](https://trello.com/c/1ENdSIfF/874-add-regioncode-to-provider-enrichment)

PR #260, opened 2019-01-24
Default created/updated to current time

timabell:
Because the zero was messing with the rails specs


timabell:
[Implement incremental fetch for providers.](https://trello.com/c/HMAlga4l/852-implement-incremental-fetch-for-providers)

PR #267, opened 2019-02-08
946 add provider last published

timabell:
### Context
The new API for UCAS Apply needs to know what to include for incremental fetches. This field will be used for that.

The test just extends the existing integration test rather than sorting out a unit test because the c# is going away anyway.

* Add Provider.LastPublishedAt
* Populate existing data with `now()`
* Make the publish button in the publish system update the date in this field.

### Tests done locally
* Ran a migration against a sanitized export from prod, new column was added and populated correctly.
* Fired up api/ui locally, checked that publish button resulted in updated `provider.last_published_at`

### Guidance to review
Run the api which will upgrade your db, check the new col is populated, poke the publish button and check it's updated.

### DIFF
--- a/src/ManageCourses.Api/Services/EnrichmentService.cs
+++ b/src/ManageCourses.Api/Services/EnrichmentService.cs
@@ -122,15 +122,18 @@ namespace GovUk.Education.ManageCourses.Api.Services
                 .OrderByDescending(x => x.Id)
                 .FirstOrDefault();
 
+            var provider = _context.Providers.Single(p => p.ProviderCode == providerCode);
+            var publishTimestamp = DateTime.UtcNow;
+            provider.LastPublishedAt = publishTimestamp;
             if (enrichmentDraftRecord != null)
             {
                 enrichmentDraftRecord.UpdatedAt = DateTime.UtcNow;
                 enrichmentDraftRecord.UpdatedByUser = userOrg.User;
-                enrichmentDraftRecord.LastPublishedAt = DateTime.UtcNow;
+                enrichmentDraftRecord.LastPublishedAt = publishTimestamp;
                 enrichmentDraftRecord.Status = EnumStatus.Published;
-                _context.Save();
                 returnBool = true;
             }
+            _context.Save();
 
             return returnBool;
         }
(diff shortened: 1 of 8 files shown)


timabell:
[Update provider when an associated record is updated](https://trello.com/c/KS8vArUK/946-update-provider-when-an-associated-record-is-updated)

src/ManageCourses.Api/Services/EnrichmentService.cs
```diff
@@ -122,15 +122,18 @@ public bool PublishProviderEnrichment(string providerCode, string email)
...
             {
                 enrichmentDraftRecord.UpdatedAt = DateTime.UtcNow;
                 enrichmentDraftRecord.UpdatedByUser = userOrg.User;
-                enrichmentDraftRecord.LastPublishedAt = DateTime.UtcNow;
+                enrichmentDraftRecord.LastPublishedAt = publishTimestamp;
                 enrichmentDraftRecord.Status = EnumStatus.Published;
-                _context.Save();
                 returnBool = true;
             }
+            _context.Save();
```
defong: So regardless if there is a draft enrichment when nothing is published for provider enrichment update provider last published at date?

src/ManageCourses.Domain/DatabaseAccess/ManageCoursesDbContext.cs
```diff
@@ -43,6 +43,8 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
             modelBuilder.Entity<Provider>()
                 .HasIndex(ui => ui.ProviderCode)
                 .IsUnique();
+            modelBuilder.Entity<Provider>()
+                .HasIndex(p => p.LastPublishedAt);
```
defong: Sure of index?

src/ManageCourses.Api/Services/EnrichmentService.cs
defong: if that is so we can backfill the last published at date, but only if there is a published enrichment

dankmitchell:
What about we doing about sites? Are we leaving that until providers can actually create and update them?

defong:
> What about we doing about sites? Are we leaving that until providers can actually create and update them?

Yes i will  presume so as there is no code to add or remove sites, the only way that a site is added or removed is via external methods.

All we really need is the database column as the placeholder.

PR #268, opened 2019-02-12
965 Rename postgres sequences to match table names

timabell:
To make it possible to set up migrations in Ruby on Rails.


timabell:
[Rename sequences to match table names](https://trello.com/c/w8PDGiVR/965-rename-sequences-to-match-table-names)

timabell:
[Move DB migrations to Rails app.](https://trello.com/c/oYqWlNMY/961-move-db-migrations-to-rails-app)

PR #270, opened 2019-02-15
[991] Replace " " main site with "-"

timabell:
To make data more consistent.
Fixes issue reported by UCAS devs for our export to apply system.

### DIFF
--- a/src/ManageCourses.UcasCourseImporter/xls-reader/XlsReader.cs
+++ b/src/ManageCourses.UcasCourseImporter/xls-reader/XlsReader.cs
@@ -236,7 +236,7 @@ namespace GovUk.Education.ManageCourses.Xls
                     var ucasCampus = new UcasCampus
                     {
                         InstCode = row.GetCell(columnMap["INST_CODE"]).StringCellValue.Trim(),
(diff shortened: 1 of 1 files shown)


timabell:
[translate campus_code from blank string to dash](https://trello.com/c/Vikb98jQ/991-translate-campuscode-from-blank-string-to-dash)

misaka:
Isn't displaying "-" for a missing campus code a concern of how we display it? I mean, if the campus code is missing, it should be nil or blank or whatever, and then when we display it we can decide how it should be represented? The bit on the ticket about "... to match other main site records." throws me, though, so maybe I'm missing wider context.

benilovj:
> Isn't displaying "-" for a missing campus code a concern of how we display it? I mean, if the campus code is missing, it should be nil or blank or whatever, and then when we display it we can decide how it should be represented? The bit on the ticket about "... to match other main site records." throws me, though, so maybe I'm missing wider context.

So `" "` is an invalid campus code – only A-Z, 0-9 and `"-"` are allowed.

In all but two cases, `" "` maps to "Main site", which in all other cases has campus code `"-"`:

image

PR #271, opened 2019-02-22
Apply campus code fixup to course data

timabell:
This was causing the in-memory join to fail to find due to the mismatch
in values.

### Context
Fire 🔥 

### Guidance to review

This is the join: https://github.com/DFE-Digital/manage-courses-api/blob/c7be25bc6b9b10ddd6b0e510311166d3c574659b/src/ManageCourses.UcasCourseImporter/xls-reader/XlsReader.cs#L72

### DIFF
--- a/src/ManageCourses.UcasCourseImporter/xls-reader/XlsReader.cs
+++ b/src/ManageCourses.UcasCourseImporter/xls-reader/XlsReader.cs
@@ -54,7 +54,7 @@ namespace GovUk.Education.ManageCourses.Xls
                         CrseTitle = row.GetCell(columnMap["CRSE_TITLE"]).StringCellValue.Trim(),
                         Studymode = row.GetCell(columnMap["STUDYMODE"]).StringCellValue.Trim(),
                         Age = row.GetCell(columnMap["AGE"]).StringCellValue.Trim(),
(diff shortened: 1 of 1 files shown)


timabell:
[missing courses](https://trello.com/c/sQHZyKVX/1026-missing-courses)

src/ManageCourses.UcasCourseImporter/xls-reader/XlsReader.cs
```diff
@@ -345,5 +345,9 @@ public List<UcasNoteText> ReadNoteText(string folder)
 
             return result;
         }
+
+        private string CampusCodeFixer(string code){
+            return code.Replace(" ", "-").Trim();
```
defong: return code.Trim().Replace(" ", "-");

defong: return code.Trim().Replace("", "-");

defong: one or both of those

timabell: optimise later

defong: lets hope `code` != buggy 

we should be a little bit more careful and null check

PR #275, opened 2019-03-08
1051 Fixed if the accrediting provider is null or empty default it to 'N'

defong:
### Context
accrediting provider is null or empty 

Fixed if the accrediting provider is null or empty default it to 'N'


defong:
[Provider API issues flagged by UCAS: null accrediting_provider](https://trello.com/c/QviFruXI/1051-provider-api-issues-flagged-by-ucas-null-accreditingprovider)

misaka:
Are there tests to go along with this?

defong:
> Are there tests to go along with this?

No, its the same vein as https://github.com/DFE-Digital/manage-courses-api/pull/270

misaka:
> > Are there tests to go along with this?
> 
> No, its the same vein as #270

That's reassuring.

defong:
> > > Are there tests to go along with this?
> > 
> > 
> > No, its the same vein as #270
> 
> That's reassuring.

That's why I requested approval from you...

Options

1. Let this be
2. Add a whole test suite for xls and csv file integration
3. Move this code further out, to find a place to test it, don't really solve 2
4. Move this to ruby

misaka:
Can we see how many providers' records will be changed by this?

I suspect it will be 0, in which case I'm happy to merge. If it's more than that I think we double-check with @benilovj to check his understanding of the data we have.

defong:
> Can we see how many providers' records will be changed by this?
> 
> I suspect it will be 0, in which case I'm happy to merge. If it's more than that I think we double-check with @benilovj to check his understanding of the data we have.

I can confirm that it is zero, so in additional to this, its seems we have rogue/orphan data as a current concern.

PR #276, opened 2019-03-11
[1013] Set course.changed_at

timabell:
To match logic for provider.changed_at

* Bring test course/enrichment codes into line now that we are testing
interaction with the course record.

### Context
Incremental api in https://github.com/DFE-Digital/manage-courses-backend/ requires this field to be set to include modified records.

### Guidance to review

:ship:


timabell:
[updated course changed_at when course published](https://trello.com/c/IfvMUVbY/1013-updated-course-changedat-when-course-published)

PR #281, opened 2019-03-25
[1133] Add manage_courses_backend__key env var to VSTS

tvararu:
### Context

This will be used by manage-courses-backend in order to trigger republishing.

Terraform prior art: https://github.com/DFE-Digital/bat-infrastructure/pull/119

### Guidance to review

I don't understand VSTS very well so don't know if adding this without going and poking the interface to define a value for `manageCoursesBackendKey` will cause everything to break.

### DIFF
--- a/azure/template.json
+++ b/azure/template.json
@@ -156,6 +156,12 @@
             "metadata": {
                 "description": "Connection string for Sentry monitoring."
             }
+        },
+        "manageCoursesBackendKey": {
+            "type": "string",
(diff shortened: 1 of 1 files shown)


tvararu:
[Config in Manage Courses Backend for bearer token auth with Manage Courses API](https://trello.com/c/i9USTN0k/1133-config-in-manage-courses-backend-for-bearer-token-auth-with-manage-courses-api)

PR #283, opened 2019-03-26
[1132] Added endpoint to be able to publish a course

defong:
### Context
Endpoint to be able to publish a course

Added `/api/Publish/internal/course/{providerCode}/{courseCode}` as a endpoint
Accepting a body ```{ "email": "string" }```

Expected return response `Bad request` or
```{  "result": true|false }```

### Guidance to review

This is almost the same as 
https://github.com/DFE-Digital/manage-courses-api/blob/0640a280f75e5d2fff0daee437432a3c772d57a7/src/ManageCourses.Api/Controllers/PublishController.cs#L37

apart lacking a call to  `_enrichmentservice.PublishCourseEnrichment(providerCode, courseCode, name);`

So when the endpoint is called it will send either a minimum course or a course using the latest existing published enrichment.

image

### CODE REVIEW

**Comment by defong** [comment]
[Manage Courses API endpoint for triggering course-republishing](https://trello.com/c/ooKtcota/1132-manage-courses-api-endpoint-for-triggering-course-republishing)

File: tests/ManageCourses.Tests/SmokeTests/BackendInternalApiEndpointTests.cs [hunk 1]
```diff
@@ -0,0 +1,121 @@
...
+    /// <summary>
+    /// These tests poke the http endpoints of the api whilst the api is running in a
+    /// a captive web host.
+    /// </summary>
+    [TestFixture]
+    [Category("Smoke")]
+    [Explicit]
+    public partial class BackendInternalApiEndpointTests : ApiSmokeTestBase
+    {
+        private const string Email = "feddie.krueger@example.org";
```
tvararu: spooky

src/ManageCourses.Api/Controllers/PublishController.cs
```diff
@@ -42,6 +44,25 @@ public async Task<ActionResult> PublishCourseToSearchAndCompare(string providerC
...
+        /// </summary>
+        /// <returns>boolean indicating success/failure</returns>
+        [HttpPost]
+        [Route("internal/course/{providerCode}/{courseCode}")]
+        [ProducesResponseType(typeof(bool), 200)]
+        [ProducesResponseType(400)]
+        [ProducesResponseType(404)]
+        [ExemptFromAcceptTerms]
+        [BackendApiTokenAuth]
+        public async Task<ActionResult> PublishCourseToSearchAndCompare(string providerCode, string courseCode, [FromBody]BackendRequest request)
```
timabell: I think we should give it a different name rather than overloading `PublishCourseToSearchAndCompare()`

How about `JustPushCourseToSearchAndCompare()`

src/ManageCourses.Api/Model/BackendRequest.cs
```diff
@@ -0,0 +1,7 @@
+namespace GovUk.Education.ManageCourses.Api.Model
+{
+    public class BackendRequest
```
timabell: add code doc to the class explaining why we need this

src/ManageCourses.Api/Controllers/PublishController.cs
defong: The need for verboseness?

src/ManageCourses.Api/Model/BackendRequest.cs
defong: More verboseness?

Its simply a dto for complex type to serve the mvc frameframe in order to understand that the body of the json that it receives needs to map to the class property.

ie everything that is under 
GovUk.Education.ManageCourses.Api.Model namespace

tests/ManageCourses.Tests/SmokeTests/BackendInternalApiEndpointTests.cs
defong: sourced from the `ApiEndpointTests` stems from all the nightmare we all have

src/ManageCourses.Api/Controllers/PublishController.cs
defong: `Publish Course To Search And Compare` vs
`Just Push Course To Search And Compare` where by the latter is fundamental the same as the former?

timabell: It's just a suggested name, the key thing is I think the overload provides opportunity for future confusion.

src/ManageCourses.Api/Model/BackendRequest.cs
timabell: It would make the PR easier to review and the code easier to understand in the future. Up to you if you're not keen.

minor.

src/ManageCourses.Api/Controllers/PublishController.cs
timabell: I'd like distinct names before I give this the thumbs up

src/ManageCourses.Api/Model/BackendRequest.cs
defong: Not keen

src/ManageCourses.Api/Controllers/PublishController.cs
defong: You can have `InternalPublishCourseToSearchAndCompare`

PR #286, opened 2019-04-01
[1225] Pass in correct enrichment status on Course model

tvararu:
### Context

This will be used in the UI when displaying the courses table, as currently there is insufficient information to decide if a course is in a draft status or not.


src/ManageCourses.Domain/Models/Course.cs
```diff
@@ -61,7 +61,7 @@ public Course()
         public IEnumerable<Site> Sites { get => CourseSites != null && CourseSites.Any() ? CourseSites.Select(x => x.Site) : null; }
 
         [NotMapped]
-        public EnumStatus? EnrichmentWorkflowStatus { get; set; }
+        public WorkflowStatus? EnrichmentWorkflowStatus { get; set; }
```
defong: Can you slap a comment, along the lines that if its null then it was not set

ie the DataService.xxx is the only method that should be setting...

Just in case we need to revisit/build on this

tvararu: Done

PR #288, opened 2019-04-04
1231 set course running

defong:
### Context

When publishing a course for an opted-in provider we need to make sure that the ucas statuses `course_site.publish` and `course_site.status`  are updated so that the course will be included in publishing to "find" and will be available on the ucas apply system.

* `course_site.publish` can be `Y`or `N`
* `course_site.status` can be new, running, discontinued or suspended (`N`,`R`,`D`,`S`)

Logic for publishing is:

* If opted-in and status is `N` or `R` set publish to `Y`
* If opted-in and status is `N` set status to `R`

Leave these fields alone for all other combinations.


defong:
[The publish button sets a new course to running](https://trello.com/c/oDsV5mO9/1231-the-publish-button-sets-a-new-course-to-running)

src/ManageCourses.Api/Services/TransitionService.cs
```diff
@@ -0,0 +1,36 @@
...
+
+            if (optedInNewCourse != null)
+            {
+                foreach(var cs in optedInNewCourse.CourseSites)
+                {
+                    cs.Publish = "Y";
+                    cs.Status = "R";
+                }
+
+                _context.Save();
```
misaka: Just to check, seeing as I'm not familiar with .NET Core, this save here will update the context but not the DB, is that right? So this change won't be saved to the DB until `_searchAndCompareService.SaveCourse(providerCode, courseCode, name);` finishes in the `PublishController`?

I'm just double-checking that if an enrichment validation fails, or any other validation really, this update won't actually be persisted to the DB.

src/ManageCourses.Api/Services/TransitionService.cs
```diff
@@ -0,0 +1,36 @@
...
+            _context = context;
+        }
+
+        public void UpdateNewCourse(string providerCode, string courseCode, string email)
+        {
+            var optedInNewCourse = _context.GetCourse(providerCode, courseCode, email)
+                .FirstOrDefault(c =>
+                    c.Provider.OptedIn &&
+                    c.CourseSites != null &&
+                    c.CourseSites.All(cs => cs.Status.Equals("n", StringComparison.InvariantCultureIgnoreCase)));
```
benilovj: So I think this mean that if a course has a mixture of `new` and `discontinued` CourseSites, it's not gonna be set to running. This means the following test will fail:

```
Given course
with site 1 which is new
and site 2 which is suspended
when 'update new course' is called
then site 1 is running and publish='y'
and site 2 is suspended
```

@fofr do you agree with the above test?

defong:
How about I just round it off as 

Given a course with at least 1 new unpublished/published site
when 'update new course' is called
then the all new unpublished sites is running and publish='y'

Given a course with no new unpublished site
when 'update new course' is called
then no change


@fofr @benilovj  @misaka

timabell:
note to self - needs a test to make sure it *doesn't* change status for non-opted-in providers (if that isn't in here somewhere already)

tests/ManageCourses.Tests/ImporterTests/DbBuilders/CourseType.cs
```diff
@@ -0,0 +1,18 @@
...
+using System;
+
+namespace GovUk.Education.ManageCourses.Tests.ImporterTests.DbBuilders
+{
+    internal enum CourseType
+    {
+        NewUnpublished,
+        NewPublished,
+        SuspensedUnpublished,
+        SuspensedPublished,
```
timabell: `s/Suspensed/Suspended/`

src/ManageCourses.Api/Controllers/PublishController.cs
```diff
@@ -38,6 +40,9 @@ public async Task<ActionResult> PublishCourseToSearchAndCompare(string providerC
         {
             var name = this.User.Identity.Name;
 
+            // update course status from new to running only if provider is opted in
```
timabell: maybe rename method or move to xml-doc on the method

src/ManageCourses.Api/Services/TransitionService.cs
```diff
@@ -0,0 +1,48 @@
...
+    public class TransitionService : ITransitionService
+    {
+        private IManageCoursesDbContext _context;
+
+        public TransitionService(IManageCoursesDbContext context)
+        {
+            _context = context;
+        }
+
+        private bool OnlyNewOrRunningAndUnpublished(CourseSite courseSite)
```
timabell: public methods first

timabell:
discussed briefly with @misaka

PR #289, opened 2019-04-09
1274 publish org endpoint

timabell:
### Context

For the new location editing that's coming up
https://manage-courses-prototype.herokuapp.com/location/-

We can't use the existing internal-publish endpoint because it
only does one course, which means site info would be stale.

The new endpoint does a subset of the work that
`PublishCoursesToSearchAndCompare()` does

This change is similar to https://github.com/DFE-Digital/manage-courses-api/commit/ca55908506fd100293793af91b42bd938f616e27#diff-88577f484b6f5060edc3361bd633642aR59

* Refactor existing backend test for brevity
* Fix assertion mistake in existing tests (`.Should().Be` instead of `Equals`)
* Renamed existing test class so that pair of test files will look right in file listing
* Add new endpoint to controller
* Duplicate existing backend regression test for new endpoint - this tests that service is actually called, which should be enough regression proofing against unknowingly breaking the endpoint in the future.

### Guidance to review

:ship:

### DIFF
--- a/src/ManageCourses.Api/Controllers/PublishController.cs
+++ b/src/ManageCourses.Api/Controllers/PublishController.cs
@@ -69,6 +69,24 @@ namespace GovUk.Education.ManageCourses.Api.Controllers
         }
 
         /// <summary>
+        /// Publishes all of a providers courses
+        /// </summary>
+        /// <returns>boolean indicating success/failure</returns>
+        [HttpPost]
+        [Route("internal/courses/{providerCode}")]
+        [ProducesResponseType(typeof(bool), 200)]
+        [ProducesResponseType(400)]
+        [ProducesResponseType(404)]
+        [ExemptFromAcceptTerms]
+        [BackendApiTokenAuth]
+        public async Task<ActionResult> InternalPublishCoursesToSearchAndCompare(string providerCode, [FromBody]BackendRequest request)
+        {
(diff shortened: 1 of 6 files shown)


src/ManageCourses.Api/Controllers/PublishController.cs
```diff
@@ -68,6 +69,24 @@ public async Task<ActionResult> InternalPublishCourseToSearchAndCompare(string p
             return Ok(new{result});
         }
 
+        /// <summary>
+        /// Publishes all of a providers courses
```
defong: providers courses where the ' ?

tests/ManageCourses.Tests/SmokeTests/ApiSmokeTestBase.cs
```diff
@@ -18,5 +23,37 @@ public void OneTimeTearDown()
         {
             LocalWebHost?.Stop();
         }
+
+        /// <summary>
+        /// Backend doesn't use the api client, but it's useful for testing so this class extends the client with methods called backend.
```
defong: grammer

timabell:
[new manage-api endpoint for re-publishing an org](https://trello.com/c/Wc4r5Fa5/1274-new-manage-api-endpoint-for-re-publishing-an-org)

PR #295, opened 2019-05-08
Add backend key to WebJob

smneal:
### Context
Fix missing manage backend key in the WebJob

Add manage backend key to WebJob

### DIFF
--- a/azure/template.json
+++ b/azure/template.json
@@ -529,6 +529,10 @@
                             {
                                 "name": "snc__api__url",
                                 "value": "[parameters('searchAndCompareAPIUrl')]"
+                            },
+                            {
+                                "name": "manage_courses_backend__key",
+                                "value": "[parameters('manageCoursesBackendKey')]"
                             }
                         ]
                     },


misaka:
> The key is in use, but this has nothing to do with web job....

So what's it for?

smneal:
> > The key is in use, but this has nothing to do with web job....
> 
> So what's it for?

The WebJob fails because it can't find the key (error:  Missing config keys: manage_courses_backend:key). If it's not needed lets remove the dependency and tidy things up.

defong:
> > > The key is in use, but this has nothing to do with web job....
> > 
> > 
> > So what's it for?
> 
> The WebJob fails because it can't find the key (error: Missing config keys: manage_courses_backend:key). If it's not needed lets remove the dependency and tidy things up.

Its the key that is used by backend to talk to api

https://github.com/DFE-Digital/manage-courses-backend/blob/66f962acfb2e91e771372842887162620cb87d9c/app/services/manage_courses_api_service.rb#L8

which web job we referring to by the way?


if its the ucas import one, that whole vertical slice can be deleted, @dankmitchell tried and failed due to emotional attachments

PR #297, opened 2019-05-29
Add asp ip to pg sql firewall whitelist

smneal:
### Context

Add resource deployment to ensure ASP outbound IP's are added to the relevant PostgreSQL firewall rules

- Add firewall rule resource deployment for API and Worker ASP

### Guidance to review


azure/template.json
```diff
@@ -572,6 +598,32 @@
...
             ]
         },
+        {
+            "apiVersion": "2017-05-10",
+            "name": "wkr-postgresql-server-firewall-rules",
+            "type": "Microsoft.Resources/deployments",
+            "properties": {
+                "mode": "Incremental",
+                "templateLink": {
+                    "uri": "[concat(variables('deploymentUrlBase'),'postgresql-server-firewall-rules.json')]",
```
defong: take it we own the template json

smneal: Yes we do. There are building blocks for all the Azure resources we're likely to use.

PR #304, opened 2019-06-05
[1583] Prevent publishing of courses with no published sites

timabell:
### Context

The code for preventing this currently lives in the sc-exporter:
https://github.com/DFE-Digital/manage-courses-api/blob/5cf6af07263c3090a8de8ff528f979756da49c8f/src/ManageCourses.CourseExporterUtil/Publisher.cs#L92

Now that the mcb tool is looping through the courses for a provider it
wasn't hitting this filter and was over-publishing to find.

Skip courses with no published sites when publishing

### DIFF
--- a/src/ManageCourses.Api/Mapping/CourseMapper.cs
+++ b/src/ManageCourses.Api/Mapping/CourseMapper.cs
@@ -16,6 +16,13 @@ namespace GovUk.Education.ManageCourses.Api.Mapping
 
         public SearchAndCompare.Domain.Models.Course MapToSearchAndCompareCourse(Domain.Models.Provider ucasProviderData, Domain.Models.Course ucasCourseData, ProviderEnrichmentModel providerEnrichmentModel, CourseEnrichmentModel courseEnrichmentModel)
         {
+            // if none of the course-sites are published, return null to prevent the whole course from being published to find
+            if (ucasCourseData?.CourseSites == null || ucasCourseData.CourseSites.All(cs =>
+                    !cs.Publish.Equals("Y", StringComparison.InvariantCultureIgnoreCase)))
+            {
+                return null; // don't allow publishing unpublished courses to find
+            }
+
             ucasProviderData = ucasProviderData ?? new Domain.Models.Provider();
             ucasCourseData = ucasCourseData ?? new Domain.Models.Course();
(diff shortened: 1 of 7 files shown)


timabell:
[prevent internal publish endpoint from publishing unpublished courses](https://trello.com/c/v4Sal6am/1583-prevent-internal-publish-endpoint-from-publishing-unpublished-courses)

src/ManageCourses.Api/Mapping/CourseMapper.cs
```diff
@@ -16,6 +16,13 @@ public class CourseMapper : ICourseMapper
 
         public SearchAndCompare.Domain.Models.Course MapToSearchAndCompareCourse(Domain.Models.Provider ucasProviderData, Domain.Models.Course ucasCourseData, ProviderEnrichmentModel providerEnrichmentModel, CourseEnrichmentModel courseEnrichmentModel)
         {
+            // if none of the course-sites are published, return null to prevent the whole course from being published to find
+            if (ucasCourseData?.CourseSites == null || ucasCourseData.CourseSites.All(cs =>
+                    !cs.Publish.Equals("Y", StringComparison.InvariantCultureIgnoreCase)))
```
defong: all courses sites need to be published?

timabell: read as "if they are all not published"
I couldn't work out how to make this read nicely in under 5 mins, I'll give it some more thought

maybe extracting a method `HasAnySitesPublished(ucasCourseData)` would help here

timabell:
might want a test for returning null when no sites published

src/ManageCourses.Api/Mapping/CourseMapper.cs
defong: https://github.com/DFE-Digital/manage-courses-backend/pull/432

i think it needs to be tackled here over this end

timabell:
map to null is already done in `MapToSearchAndCompareCourse_with_no_published_campuses`

src/ManageCourses.Api/Mapping/CourseMapper.cs
timabell: I've made the code much clearer, take another look

PR #305, opened 2019-06-10
[1609] Cope with null returned from CourseMapper

timabell:
### Context

Exporter failing

Make it not fail

To make it impossible to incorrectly publish courses the course mapper
now returns null when a course shouldn't be published. This was breaking
the calling code.

This patch makes the exporter cope with the newly returned nulls.

### Guidance to review

This code is pretty hard to test so I want to get a fix out rapidly then worry about coverage or just wait for the whole thing to be deleted.


timabell:
[course exporter failing](https://trello.com/c/euNNW7Np/1609-course-exporter-failing)

timabell:

PR #306, opened 2019-06-10
[1583] Refine failure modes for publishing unpublished courses

timabell:
### Context

* sc-exporter is falling over on newly introduced nulls (not hit prod yet)
* preview possibly broken by introducing null return

* revert change to course mapper that introduced nulls
* check for unpublished course only in publish endpoint flow

### Guidance to review

Don't ask about the tests.

### Commit message

The original patch to make the mapper return null broke other usages of
the mapper (bulk export & preview).

This second attempt operates in the publish controller instead. It still
causes attempts to publish a single course that is not published to fail
(which was the intent of the previous PR), but will leave all other code
paths untouched.

* Revert returning null in mapper
* Return null `GetValidCourses()` instead so that the publish controller
can decide what to do with non-published courses
* In the case of a request to publish a single course, fail (return
false) if null was returned from GetValidCourses.
* I think this changes the behaviour of publishing an entire provider to
only publish published courses. This might previously have been
incorrect.
* Revert tests as original mapper behaviour has been restored.
* Make sc-exporter cope with any nulls returned from course mapper


timabell:

timabell:
Approved by Misha, now add more codez....

misaka:

timabell:
abandon ship!
reverted previous commit, this no longer applies

PR #308, opened 2019-06-10
[1583] Only publish courses with running & published site(s)

timabell:
### Context

Publish endpoint(s) are publishing courses even if they have no published & running sites. The mcb command for publishing a whole provider was triggering this, which then was reset by the bulk publisher, except that the bulk publisher circuit breaker tripped when it reduced the course count back down to the correct number.

This brings the publish controller into line with the bulk publisher.

* Expand filtering in GetValidCourses to eliminate courses without any
  running/published sites (aka campuses). - this is specific to the publish endpoints unlike the original patch

### Guidance to review

You might find it easier to read each commit rather than the whole diff.

:eye: :eye: 

* The first commit is the behaviour fix, and fixup of test data.
* The second commit is a refactor to introduce a common place for the logic on the `Course` model.
* The third/fourth commits remove code that doesn't do anything.


defong:

PR #320, opened 2019-07-11
[1759] Use course.accrediting_provider_code instead of fk

timabell:
### Context

* Model from Rails side is changing to not have accrediting_provider_id,
  this is so that we can rollover a single provider without recursing.
  New column has been added by manage-courses-backend in a migration.
    
Pairing Tim+Sau

Keeping up with rails

### Guidance to review

:eyes:

### DIFF
--- a/src/ManageCourses.Api/Controllers/PublishController.cs
+++ b/src/ManageCourses.Api/Controllers/PublishController.cs
@@ -7,6 +7,7 @@ using GovUk.Education.ManageCourses.Api.Services;
 using GovUk.Education.ManageCourses.Api.Services.Publish;
 using Microsoft.AspNetCore.Mvc;
 using GovUk.Education.ManageCourses.Api.ActionFilters;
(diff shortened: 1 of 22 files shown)


timabell:
[Change course.accrediting_provider reference from id to provider_code | c#](https://trello.com/c/xtT3jjy7/1759-change-courseaccreditingprovider-reference-from-id-to-providercode-c)

timabell:
force-push - update dependency injection, remove reverse relationship

timabell:
Ready for review again having sorted out `Provider.AccreditedCourses` fallout

timabell:
Can't merge this yet. It changes the json supplied to UI, which means the accrediting provider enrichment doesn't show up. Changing that involves updating the nuget dll :cry:

timabell:
We can kick this can down the road because:

* Rails schema migration will add new cols but not delete old fk that c# depends on
* The data will not go out of sync because it is not editable by users
* The c# will be gone before the col is deleted

PR #321, opened 2019-07-12
127 investigate release failures caused by postgreserver firewall rules

rizzkhan1:
### Context
On random instances Azure DevOps release fails whilst adding firewall rules in PostgreSQL server with a conflict error message.
Both app service and worker app service attempts to add same IP address in the firewall rules list at the same time. Issue is not with the identical IP address but likely to be delay in processing the requests.

Delay worker app's postgresql firewall rules update task to start after other firewall rules update tasks.

### Guidance to review
Serialise the tasks by using dependson function

### DIFF
--- a/azure/template.json
+++ b/azure/template.json
@@ -190,7 +190,7 @@
         }
     },
     "variables": {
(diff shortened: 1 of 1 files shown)


defong (approved):

defong:
is this dead?

PR #322, opened 2019-07-15
[1728] Filter to 2019 cycle in raw sql methods

timabell:
### Context

**Making c# ~great~ work again**

The database gained a `recruitment_cycle` table, with 2019 & 2020 records from manage-courses-backend. We will soon be duplicating all data into the 2020 hierarchy of data. The c# isn't quite decommissioned yet so still needs to filter correctly to the 2019 data set. The c# *will* be gone before the 2020 cycle starts.

Pairing Tim + @defong 

* More filtering to 2019 data.
* More test coverage (changes made using TDD)

### Guidance to review

:ship: 

We've tested this with a restore from prod, editing locally in manage-courses-ui and publishing direct to search-and-compare-api. Also tested a bulk publish after further edits. :ok_hand:


timabell:
[Add recruitment cycle to provider when bulk publishing](https://trello.com/c/JmRehWVf/1728-add-recruitment-cycle-to-provider-when-bulk-publishing)

PR #327, opened 2019-08-13
1953 cleanup app settings

timabell:
### Context

We were learning dotnet core as we went, it has too many ways to configure everything.

Get as close as possible to running the tests out of the box.

To make this app easier to run and the config system easier to follow.

* Refactor - inline all the config keys
* Set default database configuration
* Moved all the defaults into `appsetting.json`.
* Made test suite read `appsettings.json`.
* Used same dev credentials as manage-courses-backend because everyone
  has mcbe which sets up this user.
* Moved config into Config folder to be more like rails.
* Expanded default config to sensible defaults.
* Stripped dev config back to only changing debug level.
* Kill mcconfig validator - more pain than gain
* Use our config builder for main app instead of implicit one from app builder.

This requires variables to be set in the azure devops web interface for signIn*

### Guidance to review

```
git clone
```

```
cd tests\ManageCourses.Tests\
dotnet user-secrets list
dotnet test
```

```
cd src/ManageCourses.Api
dotnet user-secrets list
dotnet run
```

ask me for sign-in config

See if it behaves in qa when merged.


timabell:
This is failing on DevOps because it's now running all the tests at build (including integration & smoke), and we don't currently have a postgres server on the build agent.

timabell:
[Update c# to use new SEND attribute](https://trello.com/c/DnoNVabp/1953-update-c-to-use-new-send-attribute)

